PR #24, opened 2017-11-11
feat(cli): add serve command

wbhob:
This PR implements a `nest serve` command into the Nest CLI. It is compatible both with TypeScript and JS, and is based on Nodemon.

To make this work, though, the starter projects need to be updated. Primarily, just add the `entryFile` property to `nestconfig.json`; this property should point to the main server file for Nodemon to execute.

Enjoy!

### DIFF
--- a/src/lib/handlers/serve-command.handler.ts
+++ b/src/lib/handlers/serve-command.handler.ts
@@ -33,7 +33,7 @@ export class ServeCommandHandler implements CommandHandler {
                 } else {
                     nodemon({
                         'watch': ['src/**/*.ts'],
-                        'ignore': ['src/**/*.spec.ts'],
+                        'ignore': ['src/**/*.spec.js'],
                         'exec': `ts-node ${entryFile}`
                     })
                 }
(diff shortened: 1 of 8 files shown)


src/lib/handlers/serve-command.handler.ts
```diff
@@ -0,0 +1,42 @@
...
+            .then(() => {
+                const language: string = ConfigurationService.getProperty('language');
+                const entryFile: string =
+                    path.resolve(process.cwd(),
+                        ConfigurationService.getProperty('entryFile'));
+
+                if (language == 'js') {
+                    nodemon({
+                        'watch': ['src/**/*.js'],
+                        'ignore': ['src/**/*.spec.ts'],
```
JulienDuf: `'src/**/*.spec.ts'` should be `'src/**/*.spec.js'`?

wbhob: My bad! Fixed.

wbhob: My bad! Fixed.

PR #140, opened 2018-05-26
fix(new): --skip-install never working (sprint-1)

VinceOPS:
Hi there.

I noticed that `--skip-install` did not seem to work in the `sprint-1` branch. I fixed this in `new.action.ts`. Another part of the fix is in `new.command.ts` where the property `skip-install` was read in `command`, instead of `skipInstall`. 

_NB: At the same time I took the liberty to remove the `parse` utility method which was adding more complexity than required for the creation of a simple literal object (`Input`). Let me know if I shouldn't have_ 

Starting from this point, I'd like to define a factory (or any other soft abstraction) for `*.command.ts`s and remove the sort-of duplications (filling `inputs` and `options`). If you think this is appropriate/relevant.

### DIFF
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -14,7 +14,8 @@ export class NewAction extends AbstractAction {
     const answers: Answers = await askForMissingInformation(questions);
     const args: Input[] = replaceInputMissingInformation(inputs, answers);
     await generateApplicationFiles(inputs, options);
-    if (!options.find((option) => option.name === 'skip-install')!.value) {
+    const shouldSkipInstall = options.some((option) => option.name === 'skip-install' && option.value === true);
+    if (!shouldSkipInstall) {
       await installPackages(inputs, options);
     }
   }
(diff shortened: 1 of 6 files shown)


ThomRick:
Hey @VinceOPS ! Thanks for the PR !
Unfortunately I forgot to push the skip install fix yesterday that is also fixed in your PR...
I'll merge yours because you did some nice refactoring :-)

But for the next times could you please submit issue before for project tracking.
I'm trying to make sprint development project management to have regular fix / feature releases.

I like your idea about input management in ```*.command.ts``` so don't hesitate to work on it ;-)

I'm also trying to refactor code and use Functional Programming approach because I think Object's one is not efficient enough (inheritance is not well used + usage in actions is not appropriate)

VinceOPS:
I'm not sure "FP vs OOP" is the issue. OOP is a good fit for command interfaces and everything which can be closely related to the Command pattern. But indeed there is room for improvement. 

I will catch you up very soon to make sure we are not working on the same files :+1:  !

PR #141, opened 2018-05-26
chore(debugging): allow easy debugging from VSCode

VinceOPS:
Hi there.

I think this is a missing nice-to-have: an easy way run the code with the node inspector enabled.
I ignored `.vscode` folder **except** `.vscode/launch.json` which is only intended to hold the debug configurations.
I enabled the sourcemaps generation from `tsc` so we can debug the project using the original TypeScript sources.
I added generated `.js` and `.js.map` files to `.gitignore`.

The debug is done running `*.js` files, that's why I added a `watch` npm command, in order to make the development experience smoother/nicer, as you can restart debugging after any change, since the code will always be re-compiled.  

Any feedback/critic is most welcome. 
I'd be glad to add proper documentation/README instructions if the idea is accepted :grin:


ThomRick:
Thanks for the PR @VinceOPS !

I don't like to work with the debugger but I'm not the only one contributor on this project 😛 
So I think it's a great idea because it will help developers to contribute.

BTW like your other PRs please submit issue before for project tracking 😉

ThomRick:
I think that develop a unit test is more efficient than use the debugger 😉

VinceOPS:
In my humble opinion, Testing is one thing, Debugging is an other. I agree with the fact that unit testing is the most efficient way to maintain a good working code, and that (eventually) TDD can eventually avoid a few debugging sessions. But we, as developers, often (always?) cross situations where we need to understand what's happening in the code we just wrote or we're trying to fix.

Anyway, I did not open the PR to discuss What's debugging (or what it should be) or What's testing (same) :grin: .

kamilmysliwiec:
We discussed this in 🇫🇷 already @VinceOPS. Unluckily, I have to close these PRs since they're outdated in most places 😢

PR #143, opened 2018-05-27
chore(precommit): add a precommit hook to transpile and lint

VinceOPS:
Hi there!

This is more a proposal that an enhancement or a fix: 
I think the project would benefit from a pre-commit hook which forces the new changes to respect the linter rules and to be valid TypeScript code. This is why I used `npm run lint && npm run build` as a hook. If it fails, we'll know that either the linting (tslint) or the building (tsc) stage has failed.


ThomRick:
Thanks for the @VinceOPS !

But I don't agree with that because I think that is a CI job 😉 
I explain :
If you let the possibility to manage pre-commit hook in code then developers can disable it and you loose your intend to manage the code quality.
Commit something that does not respect linting rules or unit tests is allowed. For example your can write a failing test then commit then fix it then commit again. It depends on how your are working with git.
To finish use the CI for that is better because you see that you break something (other people do also) so your are responsible to fix that you broken. It's like a game. 😉 

I said that because In my current position I installed the pre-commit hook and it was a lot of frustration then I disabled it quickly after (less than 1 week of use).

BTW let's discuss about it if you want 😉

VinceOPS:
Hi, I understand your point.
We do use this sort of hook (in my team) and, when necessary, add "-n" to `git commit` in order to bypass the hook.

I think both are not mutually exclusive (CI **and** precommit hook) but that's not a big deal!

ThomRick:
I think it could be better to use https://git-scm.com/book/gr/v2/Customizing-Git-Git-Hooks than a node module.

VinceOPS:
Well, `husky` does **exactly** that. But hooks are not shared by default... Husky is here to ensure that every single developer in the team has the hook in his working directory.

PR #186, opened 2018-09-17
fix(info): load nest dependencies versions from lock files

quen2404:
Currently, when we run `nest info`, the `[Nest Information]` section is generated from dependencies of package.json.
But if the depencency version is a range, the nest-cli only remove the symbols. For example : `"@nestjs/common": "^5.0.0"` become `common version : 5.0.0`. But this is not true, because my package manager installed a most recent version ( `@nestjs/common: 5.3.6`).
The `[Nest Information]` section should be generated from `yarn.lock` or `package-lock.json`.

### DIFF
--- a/actions/info.action.ts
+++ b/actions/info.action.ts
@@ -1,15 +1,9 @@
 import chalk from 'chalk';
-import { readFile } from 'fs';
 import { platform, release } from 'os';
 import osName = require('os-name');
-import { join } from 'path';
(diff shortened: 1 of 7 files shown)


kamilmysliwiec:
What if `yarn.lock` or `package-lock.json` doesn't exist?

jmcdo29:
Scrolling through some old PRs and issues. I think this was taken care of in #900

PR #192, opened 2018-10-17
update success message to show selected package manager after install…

raheemazeezabiodun:
After a successful installation, the `START_COMMAND` is hardcoded to `npm run start`. 
This pull request makes it dynamic based on the user input

### DIFF
--- a/lib/package-managers/abstract.package-manager.ts
+++ b/lib/package-managers/abstract.package-manager.ts
@@ -11,7 +11,7 @@ import { ProjectDependency } from './project.dependency';
 export abstract class AbstractPackageManager {
   constructor(protected runner: AbstractRunner) {}
 
-  public async install(directory: string) {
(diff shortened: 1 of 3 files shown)


lib/package-managers/abstract.package-manager.ts
```diff
@@ -11,7 +11,7 @@ import { ProjectDependency } from './project.dependency';
 export abstract class AbstractPackageManager {
   constructor(protected runner: AbstractRunner) {}
 
-  public async install(directory: string) {
+  public async install(directory: string, selectedPackageManager: string) {
```
kamilmysliwiec: maybe just `packageManager: string` (I think it is pretty self-describable in this context)

raheemazeezabiodun: Okay, I will update it

Issue #232, opened 2019-03-07
Provide user the option to skip git initialization.


METACEO:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
Running `nest new projectName` will create a new project and will initialize a git repository.

## Expected behavior
Like the Angular CLI, it would be great to offer the user the ability to skip the git initialization. From the Angular CLI `--help` output:
```
  --skip-git (-g)
    When true, does not initialize a git repository.
```

## Minimal reproduction of the problem with instructions
1. Create a new project with the Nest CLI.
2. See the `.git` folder.

## What is the motivation / use case for changing the behavior?
If I'm generating a new Nest project within a monorepo (where I already have a git project initialized) then I do not want the Nest CLI to create a new git repo within the one I already have. 

## Environment
Within an existing repository.

johannesschobel:
i see your point, yeah.. but you can easily remove the `/path-to-your-nest-application/.git` folder, right?! ;)

johannesschobel:
maybe you can submit a PR for the upcoming `@nestjs 6.0` version, which is just around the corner (note that the `6.0-rc1` has been released 2 days ago!)

METACEO:
@johannesschobel yea, I'll give it a try!

kamilmysliwiec:
Looking forward @METACEO :) Thanks for reporting

METACEO:
I took awhile @kamilmysliwiec but I've opened a very small PR.

kamilmysliwiec:
🔥 Thank you @METACEO. Published as a new minor version!

PR #265, opened 2019-04-06
Add a flag and check to skip git repository initialization.

METACEO:
In response to the #232 request, we're providing the user the option to skip git initialization.


kamilmysliwiec:
Amazing. Thank you!

PR #319, opened 2019-06-22
remove double-space in package manager question

apricote:
I removed a double space in the 
> Which package manager would you :heart: to use?

message that is shown upon initialization of a new project.

Other strings in the file also contained double spaces after emojies, but the emoji was at the beginning of the string, so it might have been an intentional styling. I left them as they were before.


kamilmysliwiec:
Thank you!

PR #378, opened 2019-09-23
Make schematic runner more robust. 

x3cion:
Now working on Windows and with local nest-cli installations over npx.
- Changed finding logic: Now using the modules resolution system to test the usual require locations
- Changed tests: Now mocking existsSync by checking for rightPath


kamilmysliwiec:
Thank you!

PR #379, opened 2019-09-23
Prettier and linter would differ in configuration

x3cion:
- Prettier: Preserve quotes
- package.json: Make tslint load the npm way
- nest.ts, generate.command.ts, gulpfile.ts: Prettier changes


kamilmysliwiec:
Thank you!

PR #385, opened 2019-09-26
feat(prompts): prompt for project selection with generate

MarkPieszak:
image


kamilmysliwiec:
```
FAIL test/lib/configuration/nest-configuration.loader.spec.ts
```

it seems that unit tests are failing now

lib/configuration/defaults.ts
```diff
@@ -5,6 +5,8 @@ export const defaultConfiguration: Required<Configuration> = {
   sourceRoot: 'src',
   collection: '@nestjs/schematics',
   entryFile: 'main',
+  projects: {},
+  monorepo: false,
```
kamilmysliwiec: I'm not entirely sure if setting `projects` to `{}` is safe (previously, this value was `undefined` so there might be a check somewhere that compares this value explicitly to `undefined`)

actions/generate.action.ts
```diff
@@ -31,10 +37,44 @@ const generateFiles = async (inputs: Input[]) => {
...
+        defaultProjectName = property + defaultLabel;
+      }
+    }
+
+    // Re-order projects to make sure Default is at the top
+    let projects: string[] = Object.keys(configurationProjects);
+    if (configuration.sourceRoot !== 'src') {
+      projects = projects.filter(p => p !== defaultProjectName.replace(defaultLabel, ''));
+    }
+    projects.unshift(defaultProjectName);
```
kamilmysliwiec: We should extract 63-68 to a separate, pure function

actions/generate.action.ts
```diff
@@ -31,10 +37,44 @@ const generateFiles = async (inputs: Input[]) => {
...
     ? getValueOrDefault(configuration, 'sourceRoot', appName)
     : configuration.sourceRoot;
+
+  // If you only add a `lib` we actually don't have monorepo: true BUT we do have "projects{}"
+  // Ensure we don't run for new app/libs schematics
+  if (
+    ['app', 'sub-app', 'library', 'lib'].indexOf(schematic) &&
+    configurationProjects &&
+    Object.entries(configurationProjects).length === 0 &&
+    !appName
```
kamilmysliwiec: Let's extract this condition to a separate function

lib/configuration/defaults.ts
MarkPieszak: 🐈

kamilmysliwiec:
💪

PR #391, opened 2019-09-30
updated  command shown on the cli at startup

rubiin:
Updated the startup command shown on the console after the new application was successfully created


rubiin:
@kamilmysliwiec  have a look at this

PR #398, opened 2019-10-03
Replacing `[PACKAGE_MANAGER] run start`

theghostyced:
## What Does This PR Do?

> This PR basically replaces the start command message from `[PACKAGE_MANAGER] run start` to `[PACKAGE_MANAGER] start` since it is faster to type and both supported package managers `i.e yarn and npm` both support this command 😁


kamilmysliwiec:
This will create unnecessary confusion for the newcomers. For instance, why `npm start` works, while `npm start:dev` doesn't? Let's stick to the `npm run` convention

Issue #420, opened 2019-10-22
Is NODE_ENV replaced by string 'none' during nest start execution!


iamsteeve:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ x ] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

It is exactly a problem that affects projects that use NODE_ENV. The problem is that when changing the project to monorepo mode and when executing npm run start: dev (nest start --watch), NODE_ENV is filled in with a string that says "none" during execution instead of being undefined and cannot be replaced

## Expected behavior

NODE_ENV = undefined
It is an important environment variable since it indicates which environment to run.

## Minimal reproduction of the problem with instructions

1.- nest new project
2.- cd project
3.- nest g lib integrations
4.- npm run start:dev (nest start --watch)
NODE_ENV === 'none'
## What is the motivation / use case for changing the behavior?

Without NODE_ENV I do not execute the project :c and I am integrating monorepo.

## Environment

```
[System Information]
OS Version     : macOS Catalina
NodeJS Version : v10.16.3
NPM Version    : 6.12.0
[Nest Information]
platform-express version : 6.7.2
common version           : 6.7.2
core version             : 6.7.2
```

iamsteeve:
https://github.com/webpack/webpack/issues/7470
nodeEnv?
I think this is the problem, xD although I have no idea if it is the cli or core of nest xD I will clone the projects in any case to verify xD (excuse my English I am Mexican :p)

kamilmysliwiec:
Fixed in 6.11.0

PR #421, opened 2019-10-22
Fix webpack config - NODE_ENV had a string 'none'

iamsteeve:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x ] Other... Please describe: It is a configuration detail
```

I guess more than one user will have got stuck with this detail, and I think the NODE_ENV should be a pure variable. The value 'undefined' must be per base, I think it has the same effect replacing the default setting but I feel it must be so.

Issue Number: #420 

## What is the new behavior?
NODE_ENV === undefined
Now NODE_ENV will be undefined by nature and can be replaced.

```
[ ] Yes
[ X ] No
```

### DIFF
--- a/lib/compiler/defaults/webpack-defaults.ts
+++ b/lib/compiler/defaults/webpack-defaults.ts
@@ -51,6 +51,9 @@ export const webpackDefaultsFactory = (
     ],
   },
   mode: 'none',
+  optimization: {
+    nodeEnv: false,
+  },
   plugins: [
     new webpack.IgnorePlugin({
       checkResource(resource: any) {


Themandunord:
👍

iamsteeve:
```js
  return {
    ...options,
    optimization: {
      nodeEnv: false,
    },
  };
}

```
Using a webpack.config.js with the indicated configuration can have the same result, although my position will always be that the CLI should leave NODE_ENV, in Development mode, NODE_ENV could not be used without adding extra webpack configuration.
Would it be indicated to write this in the documentation as if it were an option?

kamilmysliwiec:
Thank you!

Issue #314, opened 2019-06-14
Disable spec generation for specific files in nest-cli.json


rweisse:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

At the moment I have to add `--no-spec` to the `nest generate` command for specific files I don't want to be generated spec-files for.

## Expected behavior

I would prefer to be able to configure this behaviour by `nest-cli.json` file. For example in this way:
```
"spec": {
    "class": false,
    "component": true,
    "module": false,
    "service": true
  }
```
Maybe this is already possible and I have missed this in the documentation?

## What is the motivation / use case for changing the behavior?

Laziness ;-)

Zacknero:
@rweisse but witch controller don't work :(

utsav-upendra:
Hey has anyone picked this up? I would like to work on this if no-one already is.

kamilmysliwiec:
Feel free to proceed with this @utsav-upendra :)

egodsk:
Should be closed as PR #527 has been merged?

kamilmysliwiec:
You're right. Thanks for noticing @egodsk!

klachhab:
Just add to nest-cli.json : 
"generateOptions": {
    "spec": false
}

PR #527, opened 2020-01-09
feat(cli) Add generateOptions to configuration

egodsk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently you must specify `--no-spec` on *ALL* generate commands, to not pass the `--spec` flag onwards to the collection, that generates the files. No configuration is available to allow for the `--no-spec` command to be default behavior, unless otherwise specified.

This PR fixes/closes:
- #314 

## What is the new behavior?
With this feature, you are able to specify a `generateOptions` in the configuration file, which currently only have the ability to change the generation options for `spec`, however you can then specify `spec` to be either a boolean (true/false) or an object, which the different generate names as key, with booleans as value, like so:
```
"generateOptions": {
  "spec": true
}
```
or
```
"generateOptions": {
  "spec": {
    "service": true
    ...
  }
}
```

The arguments for generate has been slightly altered, to allow for this change. There has been added a new `--spec <true/false>` option, such that the `spec` parameter is not default to true when the command is passed to the action, but undefined.

This feature follows a hierarchy, such that the CLI arguments have highest priority, then comes the projects individual options (options specified within the `"projects": {}...` option), and lastly the overall `generateOptions` for the project.

Highest:
```
nest g service Cats --spec true
```

Lower:
```
{
  "sourceRoot": "src",
  "projects": {
    "cat-project": {
      "generateOptions": {
        "spec": true
      }
    }
  }
  ...
}
```

Lowest:
```
{
  "sourceRoot": "src",
  "generateOptions: {
   "spec": true
  }
  ...
}
```

Such that if `--spec true` is set on the CLI and the `spec` in the `nest-cli.json` options is set to false, it would overwrite and pass true to the action, and so forward.

One thing I’m unable to currently figure out, is that you can only use the generateOptions with the spec as an object if you use the full name of the schematic you want to run. Ex `service` and not `s` - if anyone has a good suggestion on how to be able to handle aliases, please let me know! :) 

```
[ ] Yes
[ ] No
[x] Maybe?
```
The generate command, has been changed such that the `spec` argument is no longer default to true, but undefined, unless otherwise specified in the CLI arguments. However for the `nestjs/schematics`, it looks like the spec default is always true? Please inform me if otherwise / if needs to be changed

Make sure that if your custom collection needs `--spec` to be passed as default to the action, that you in the future write `--spec true` or add it to your `nest-cli.json` configuration file

Looking forward to comments and feedback :)
- Emil G

PS. I'm not sure where to add docs to this feature? If you tell me, I will try and write some :)

### DIFF
--- a/lib/configuration/configuration.ts
+++ b/lib/configuration/configuration.ts
@@ -15,6 +15,10 @@ interface CompilerOptions {
   deleteOutDir?: boolean;
 }
 
+interface GenerateOptions {
+  spec?: boolean | Record<string, boolean>;
+}
+
 export interface ProjectConfiguration {
   type?: string;
   root?: string;
@@ -31,6 +35,7 @@ export interface Configuration {
   entryFile?: string;
   monorepo?: boolean;
   compilerOptions?: CompilerOptions;
+  generateOptions?: GenerateOptions;
   projects?: {
     [key: string]: ProjectConfiguration;
   };
--- a/actions/generate.action.ts
+++ b/actions/generate.action.ts
@@ -13,6 +13,7 @@ import {
   askForProjectName,
   moveDefaultProjectToStart,
   shouldAskForProject,
+  shouldGenerateSpec,
 } from '../lib/utils/project-utils';
 import { AbstractAction } from './abstract.action';
 
@@ -30,6 +31,7 @@ const generateFiles = async (inputs: Input[]) => {
     .value as string;
   const appName = inputs.find(option => option.name === 'project')!
     .value as string;
+  const spec = inputs.find(option => option.name === 'spec');
 
   const collection: AbstractCollection = CollectionFactory.create(
     collectionOption || configuration.collection,
@@ -43,6 +45,16 @@ const generateFiles = async (inputs: Input[]) => {
   let sourceRoot = appName
     ? getValueOrDefault(configuration, 'sourceRoot', appName)
     : configuration.sourceRoot;
+
+  const specValue = spec!.value as boolean;
+  const specOptions = spec!.options as any;
+  let generateSpec = shouldGenerateSpec(
+    configuration,
+    schematic,
+    appName,
+    specValue,
+    specOptions.passedAsInput,
+  );
 
   // If you only add a `lib` we actually don't have monorepo: true BUT we do have "projects"
   // Ensure we don't run for new app/libs schematics
@@ -74,9 +86,21 @@ const generateFiles = async (inputs: Input[]) => {
     if (project !== configuration.sourceRoot) {
       sourceRoot = configurationProjects[project].sourceRoot;
     }
+
+    if (answers.appName !== defaultProjectName) {
+      // Only overwrite if the appName is not the default- as it has already been loaded above
+      generateSpec = shouldGenerateSpec(
+        configuration,
+        schematic,
+        answers.appName,
+        specValue,
+        specOptions.passedAsInput,
+      );
+    }
   }
 
   schematicOptions.push(new SchematicOption('sourceRoot', sourceRoot));
+  schematicOptions.push(new SchematicOption('spec', generateSpec));
   try {
     const schematicInput = inputs.find(input => input.name === 'schematic');
     if (!schematicInput) {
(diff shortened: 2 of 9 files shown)


kamilmysliwiec:
Amazing contribution, thank you!

One thing before I dive into the code: would it be possible to avoid introducing breaking changes? I'm referring to this change:

"The generate command, has been changed such that the spec argument is no longer default to true, but undefined, unless otherwise specified in the CLI arguments."

Also, 2 questions:

1. Is `--no-spec` still supported?
2. Can we treat `--spec` (without `true`) same as `--spec true` (less typing, pretty much the same meaning)

> PS. I'm not sure where to add docs to this feature? If you tell me, I will try and write some :)

Here's the docs repo https://github.com/nestjs/docs.nestjs.com. We should update both CLI->Usage (https://docs.nestjs.com/cli/usages, `nest generate` options section) and CLI->Workspaces chapters (https://docs.nestjs.com/cli/monorepo, new section?)

egodsk:
Hi @kamilmysliwiec 

Thank you for your feedback!

I have looked back at my PR, and changed a couple of things related to your comment. First off to answer your questions:
> 1. Is --no-spec still supported?
Of course! you can either supply `--no-spec` or `--spec` to the CLI in order to overwrite any `generateOptions`

> 2. Can we treat --spec (without true) same as --spec true (less typing, pretty much the same meaning)
I have changed `--spec` to now support `--spec` *ONLY*, and no longer with either true or false (Use `--spec` for true and `--no-spec` for false).

I was not able to find a solution which allowed for both the shorthand `--spec` as well as `--spec [true/false]` (Believe me i tried), so I removed the ability to use `--spec [true/false]`, which kind of only also makes the cli cleaner in a way :) 

In regards to your comment about breaking changes:
> One thing before I dive into the code: would it be possible to avoid introducing breaking changes? I'm referring to this change:
>
> "The generate command, has been changed such that the spec argument is no longer default to true, but undefined, unless otherwise specified in the CLI arguments."

I have found and implemented a solution which now again supports the default of true, such that no breaking changes should be in this PR! Hurra!

The way I handle this, is by generating a custom object in the case of the `--spec` command being specifically specified in the CLI, such that we know when `spec` was specified in the CLI or if it was just assigned to the default value (Could not find a way to see via the Commander.js library, so implemented it this way instead).

egodsk:
Have added a PR on the docs repo: https://github.com/nestjs/docs.nestjs.com/pull/931

egodsk:
Quick rebase on nestjs/nest-cli to resolve conflicts :)

lib/configuration/configuration.ts
```diff
@@ -15,6 +15,14 @@ interface CompilerOptions {
   deleteOutDir?: boolean;
 }
 
+interface GenerateOptions {
+  spec?:
+    | boolean
+    | {
+        [key: string]: boolean;
+      };
```
kamilmysliwiec: What about using `Record<string, boolean>` instead?

egodsk: I guess it could be used instead, however I don’t really see the real benefit of changing it to that instead? Could you explain?

kamilmysliwiec: This type is predefined and comes with TypeScript (so there is no reason to define a new one) + we're using `Record` a lot in all Nest packages, hence, it would be nice to stay cohesive

egodsk: Ah! Yes, will commit a change for this later.

kamilmysliwiec: Thank you :)

kamilmysliwiec:
Thank you! Amazing contribution :)

John0x:
This has not yet been released on npm right?
Just making sure, cause it ain't working with nest cli 6.14.0

egodsk:
Hi @John0x

Do you mean there is a bug in the code, or that it does not seem like it has been released on npm?

abigmiu:
when i set `spec` as `false`,  but i use `nest g res` it still generate spec file 

image

Haze272:
This feature still doesn't work could we see it in nest updates?

PR #607, opened 2020-03-16
feat(cli): Added inspect-brk flag

atjeff:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: https://github.com/nestjs/nest-cli/issues/592

## What is the new behavior?
New flag debug-brk to pass inspect-brk with port or by itself

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

PR #643, opened 2020-03-27
Revert "feat(cli): Added inspect-brk flag"

kamilmysliwiec:
Reverts nestjs/nest-cli#607


maverick1872:
Curious why this was reverted? Was hoping to update my service's debug script to leverage this but it appears that it's not supported due to this revert.

PR #788, opened 2020-07-10
feat(cli): add pnpm package manager support

dkantereivin:
## PR Checklist

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

nest-cli only offers `npm` and `yarn` as package managers. As `pnpm` is growing rapidly into a third major player in the JavaScript package manager ecosystem, nest-cli should support it both when running `nest new` and throughout for running nest scripts and installing dependencies.

## What is the new behavior?
pnpm is now available as a package manager, like npm and yarn. Since pnpm uses the same commands [requires no significant changes](https://pnpm.js.org/en/pnpm-vs-npm) compared to working with npm, there are no substantial changes to codebase other than the addition of new objects/references for pnpm (and one logic change). pnpm passes all tests, including when using nest-cli to run the default app using pnpm-installed packages.
image

- [ ] Yes
- [x] No

## Other information
NOTE: Nest Official Documentation still requires updating [here](https://docs.nestjs.com/cli/usages). Please see docs PR [here](https://github.com/nestjs/docs.nestjs.com/pull/1347)

### DIFF
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -174,6 +174,7 @@ const askForPackageManager = async (): Promise<Answers> => {
     generateSelect('package-manager')(MESSAGES.PACKAGE_MANAGER_QUESTION)([
       PackageManager.NPM,
       PackageManager.YARN,
+      PackageManager.PNPM
     ]),
   ];
   const prompt = inquirer.createPromptModule();
--- a/lib/configuration/defaults.ts
+++ b/lib/configuration/defaults.ts
@@ -26,6 +26,7 @@ export const defaultGitIgnore = `# compiled output
 logs
 *.log
 npm-debug.log*
+pnpm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
--- a/lib/package-managers/index.ts
+++ b/lib/package-managers/index.ts
@@ -3,5 +3,6 @@ export * from './package-manager.factory';
 export * from './abstract.package-manager';
 export * from './npm.package-manager';
 export * from './yarn.package-manager';
+export * from './pnpm.package-manager';
 export * from './project.dependency';
 export * from './package-manager-commands';
--- a/lib/package-managers/package-manager.factory.ts
+++ b/lib/package-managers/package-manager.factory.ts
@@ -3,6 +3,7 @@ import { AbstractPackageManager } from './abstract.package-manager';
 import { NpmPackageManager } from './npm.package-manager';
 import { PackageManager } from './package-manager';
 import { YarnPackageManager } from './yarn.package-manager';
+import { PnpmPackageManager } from './pnpm.package-manager';
 
 export class PackageManagerFactory {
   public static create(name: PackageManager | string): AbstractPackageManager {
@@ -11,6 +12,8 @@ export class PackageManagerFactory {
         return new NpmPackageManager();
       case PackageManager.YARN:
         return new YarnPackageManager();
+      case PackageManager.PNPM:
+        return new PnpmPackageManager();
       default:
         throw new Error(`Package manager ${name} is not managed.`);
     }
@@ -22,8 +25,12 @@ export class PackageManagerFactory {
         if (error) {
           resolve(this.create(PackageManager.NPM));
         } else {
-          if (files.findIndex(filename => filename === 'yarn.lock') > -1) {
+          if (files.findIndex((filename) => filename === 'yarn.lock') > -1) {
             resolve(this.create(PackageManager.YARN));
+          } else if (
+              files.findIndex((filename) => filename === 'pnpm-lock.yaml') > -1
+          ) {
+            resolve(this.create(PackageManager.PNPM));
           } else {
             resolve(this.create(PackageManager.NPM));
           }
--- a/lib/package-managers/package-manager.ts
+++ b/lib/package-managers/package-manager.ts
@@ -1,4 +1,5 @@
 export enum PackageManager {
   NPM = 'npm',
   YARN = 'yarn',
+  PNPM = 'pnpm'
 }
--- a/lib/package-managers/pnpm.package-manager.ts
+++ b/lib/package-managers/pnpm.package-manager.ts
@@ -1,0 +1,27 @@
+import { Runner, RunnerFactory } from '../runners';
+import { PnpmRunner } from '../runners/pnpm.runner';
+import { AbstractPackageManager } from './abstract.package-manager';
+import { PackageManager } from './package-manager';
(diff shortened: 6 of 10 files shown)


jmcdo29:
@kamilmysliwiec I've cloned and tested this myself. Seems to work great from what I can tell :+1:

dkantereivin:
Update: I've been using this for several days now to run a Nest application I've been working on, and so far I've not run into any issues.

dkantereivin:
@kamilmysliwiec would appreciate if this could be reviewed when you have a moment - my team hopes to use nest-cli and pnpm on our new project, and would ideally like to get cli updates as they come. Thanks so much!

jmcdo29:
@kamilmysliwiec could this be a part of the next release?

kamilmysliwiec:
What's the benefit in introducing a new option in the `nest new` command? Generally, it doesn't matter what package manager you use with Nest. Currently, if you want to use `pnpm`, you have to run the following commands:

```typescript
$ nest new APP_NAME --skip-install
$ pnpm install
```

And so imo, adding a new option to the CLI brings more confusion to the new framework users, especially if they are new to the Node.js ecosystem. We should leave this decision to the more experiencedusers instead of overwhelming them with all the available options.

dkantereivin:
I hear you, but I don't really see it as another option, at least not one that adds complexity. I would agree that adding config options adds layers of complexity, but when it comes to choosing a package manager, newer users simply look at the screen and go "I'm using NPM, I see NPM on the screen, so that must be it," and proceed with selecting that option. Much in the same way that new users would ignore the yarn option if they use npm, because they don't know what it is. If you don't feel it's useful as it's own option in the dialogue menu, I would hope it could at least be supported through `--package-manager pnpm`. I can make this adjustment to the PR if you'd like.

Your argument that users can just run `pnpm install` (and delete `package-lock.json`) applies to yarn too - they could use npm, delete `package-lock.json` and run `yarn install`, and they'll be presented with a shiny new project using yarn. But since CLIs are meant to provide ease of use for many use cases (which is why the command line is so powerful, after all), nest-cli handles this step for you through the yarn option. And if you're already going to make users choose a package manager in the installation screen, you might as well present them with the option to use their package manager.

As well, even though nest-cli doesn't really use the package manager internally (as the cli doesn't handle dependencies), pnpm users unfamiliar with the framework (or with pnpm) might get the impression that pnpm is **not compatible/supported** by the cli, due to some conflicts - this is of course not the case, but I can understand users might assume this.

Ultimately, I still think this PR is beneficial and is true to the purpose of a CLI: making boilerplate a breeze for a wide variety of users and use cases (including the pnpm ones!). Given that this change:
 - has very low maintenance cost since pnpm mirrors the npm interface for all functions in the PR
- doesn't introduce complexity for those not using it, since they can simply not click it (as they do with npm/yarn)
- increases clarity and ease of use for anyone who knows of pnpm, and especially those who use it
- integrates pnpm, as several other CLIs are doing (including Vue Cli)
This PRs benefits outweigh the potential downsides you've outlined - plus, the CLi is platform agnostic for a reason!

P.S. sorry for the text wall there

kamilmysliwiec:
Thanks for the additional explanation! I appreciate the feedback.

> Your argument that users can just run pnpm install (and delete package-lock.json) applies to yarn too - they could use npm, delete package-lock.json and run yarn install, and they'll be presented with a shiny new project using yarn

I've never said that they would have to delete `package-lock.json` file. If you follow commands that I posted, no file would be generated and so the user would have an empty boilerplate code and can proceed with any package manager that exists (without any additional actions) in the Node.js ecosystem.

> has very low maintenance cost since pnpm mirrors the npm interface for all functions in the PR and nest-cli provides cleanly abstracts package managers

Only for now. We can't anticipate the future of pnpm (as well as other package managers). Adding more integrations out-of-the-box makes it harder to maintain the package over time. We already had some issues with Yarn2 and since Yarn is an option available in the CLI, developers assumed that we **must** align to breaking changes introduced by their team. This can happen with pnpm too.

> increases clarity and ease of use for anyone who knows of pnpm, and especially those who use it

And decreases clarity and brings confusion to newcomers and anyone switching from a different language/platform to JS & Node.js.

Perhaps I'm a little biased. WIll spend some time thinking about this and will get back to this PR shortly.

andreialecu:
I think a middle ground solution here would be to add an option in the UI to skip install. `skip-install` is even less discoverable and not really an option by itself IMO.

```
Which package manager do you want to use?
[ ] npm
[ ] yarn
[ ] Other/I will decide later
```

dkantereivin:
Yeah, you could add other, but I don't think that's adding "less complexity" than PNPM especially since there's not that much out there.

To respond to Kamil's note about not being able to anticipate the future of pnpm, that's true, however I think the risk of pnpm adding to maintenance type is very low, and especially lower than npm, since a big idea behind pnpm is that it does not change the npm interface whatsoever (or at least, all npm behaviors are identical on pnpm from the end user perspective). This seems very unlikely to change, since it's a key selling point for pnpm.

I think the demand for this, while perhaps small, is significant and is growing.

andreialecu:
FWIW, I'm a `yarn` user and I have been looking for the option to skip installation every time I bootstrapped a Nest project. I have to `ctrl+c` out of installation most times, because I already know I want to either:

1) Install additional packages right from the start (such as `@nestjs/graphql`, etc)
2) Am using `yarn` workspaces and I'd rather run install from the root, instead of the package folder.
3) I want to switch to `yarn` version 2.0 prior to running the install.

TemaSM:
Any updates here? pnpm is really good option as package manager.
+1 for adding pnpm by merging this PR

jmcdo29:
@kamilmysliwiec what do you think of reviewing this again? Personally, I am a `pnpm` user, it's stable and doesn't seem like it's going to be changing anytime soon. It's CLI is pretty much just like `npm`'s with some additional commands that don't need to be used by default (certainly not for dependency install). 

For the meantime, I've been using `nest new <project> --skip-install && cd <project> && pnpm i` and while it works, it would be nice to just choose pnpm by default. I'm cool either way, but figure we should make a decision and either merge this PR or get it closed out.

qfox:
Would be great to move skip-install option to global config file. It allows to skip passing skip-install option each time

dkantereivin:
Happy anniversary to this PR 🥳

kamilmysliwiec:
LGTM

Thanks for your contribution & patience

Issue #868, opened 2020-09-03
Webpack compiler does not display e.g file system permission errors


leptoquark1:
I'd the case that the root folder had insufficient permission, resulting in an js error when the error obj was set but not checked.

https://github.com/nestjs/nest-cli/blob/d88e4940d995d2907fc14264307945375b685b2d/lib/compiler/webpack-compiler.ts#L78

err can be set but stats undefined.

kamilmysliwiec:
Would you like to create a PR for this issue?

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest-cli/pull/884

PR #884, opened 2020-09-16
fix(): fatal errors during build won't prompt (webpack)

leptoquark1:
## PR Checklist

- [-] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The err object was not checked and it was relied on defined stats. This results in a confusing type error that has nothing to do with the actual error.

Issue Number: #868 

## What is the new behavior?

Now early exit when err object is set but stats remains undefined.

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/compiler/webpack-compiler.ts
+++ b/lib/compiler/webpack-compiler.ts
@@ -75,6 +75,13 @@ export class WebpackCompiler {
     const compiler = webpack(webpackConfiguration);
 
     const afterCallback = (err: Error, stats: any) => {
+      if (err && stats === undefined) {
+        // Could not complete the compilation
+        // The error caught is most likely thrown by underlying tasks
+        console.log(err);
+        return process.exit(1);
+      }
+
       const statsOutput = stats.toString({
         chunks: false,
         colors: true,


lib/compiler/webpack-compiler.ts
```diff
@@ -75,6 +75,11 @@ export class WebpackCompiler {
     const compiler = webpack(webpackConfiguration);
 
     const afterCallback = (err: Error, stats: any) => {
+      if (err && stats === undefined) {
+        console.log(err);
+        return process.exit(1);
```
kamilmysliwiec, line 80: Do we want to close the current process even if the `watch` mode is turned on?

leptoquark1: Since the err object is detached from the actual compiler result and issues an error catched by webpack which is most likely thrown by underlying tasks, it should be handled fatal.

In my test scenarios this were mostly 

```json
{
  errno: -13,
  code: 'EACCES',
}
```

kamilmysliwiec: Can we add an inline comment then? To explain why this if () block is needed

leptoquark1:
@kamilmysliwiec I resolved your review. Any other suggestions?

kamilmysliwiec:
Thanks!

Issue #896, opened 2020-09-30
Read package versions from lock file instead of package.json


jmcdo29:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Right now, `nest info` reads from `package.json` and outputs the versions set in there. This could mean that `nest info` prints `common: 7.0.0` even when the latest version is installed (this can be seen with `nest new test-project && cd test-project && nest info`)

## Expected behavior

`nest info` should read from the appropriate lock file and print out the version of the currently installed nest packages.

## Minimal reproduction of the problem with instructions

```sh
nest new test-project
cd test-project
nest info
```

## What is the motivation / use case for changing the behavior?

It would be nice to be able to get current information on the installed packages instead of what the package.json says, which _could_ be less accurate than the lockfile. Also, I've had several people be confused on what version is installed when using `nest info` because it doesn't read from the lock file. 

## Environment

```
[System Information]
OS Version     : Linux 5.4
NodeJS Version : v12.18.2
NPM Version    : 6.14.5 

[Nest CLI]
Nest CLI Version : 7.5.1 

[Nest Platform Information]
platform-express version : 7.0.0
common version           : 7.0.0
core version             : 7.0.0
```

## Other Information

I'm willing to work on a PR for this, actually kind of excited to work more with the lock files. However, would we be supporting the major three package managers (npm, yarn, and pnpm) or just the two most popular (npm and yarn). Either way, a yml parse will need to be used for yarn.lock files

jmcdo29:
If we're going to support PNPM (which I hope we do) we should merge https://github.com/nestjs/nest-cli/pull/788. For now, I'll create a PR that works with NPM and Yarn.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest-cli/pull/900

PR #900, opened 2020-10-02
feat: reads versions from lock files

jmcdo29:
Reading from the appropriate lock files instead of the
package.json means having a better idea of what packages
are installed. With the package.json, there's a chance to
miss the actual installed version due to using a range
like ^7.0.0 whereas with the lock file, the version is
printed to what the package manager resolved.

To make this more easily achievable, I refactored the
info.action.ts file to be more class based so I could
reuse the package manager instead of having to re-find it
each time.

re: #896

## PR Checklist

Tests didn't already exist, though I'm happy to try and write some up

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently we are reading the nest package versions from the `package.json`. 
Issue Number: #896 

## What is the new behavior?
After this PR we will read them from the appropriate lock file.

```
[ ] Yes
[ ] No
```

Kind of depends on your definition of breaking. Though it should be noted that this PR doesn't support PNPM as we need to merge #788 for that. Because of this, PNPM users will lose access to `nest info` until that can be taken up, **or** we have the old functionality still in place (for pnpm and for not finding a lock file).

### DIFF
--- a/actions/info.action.ts
+++ b/actions/info.action.ts
@@ -45,30 +45,87 @@ export class InfoAction extends AbstractAction {
     console.info('NodeJS Version :', chalk.blue(process.version));
     await this.displayPackageManagerVersion();
   }
-};
 
-const displayCliVersion = () => {
-  console.info(chalk.green('[Nest CLI]'));
-  console.info(
-    'Nest CLI Version :',
-    chalk.blue(require('../package.json').version),
-    '\n',
-  );
-};
+  async displayPackageManagerVersion() {
+    try {
+      const version: string = await this.manager.version();
+      console.info(`${this.manager.name} Version    :`, chalk.blue(version), '\n');
+    } catch {
+      console.error(`${this.manager.name} Version    :`, chalk.red('Unknown'), '\n');
+    }
+  }
 
-const readProjectPackageJsonDependencies = async (): Promise<
-  PackageJsonDependencies
-> => {
-  return new Promise<PackageJsonDependencies>((resolve, reject) => {
-    readFile(
-      join(process.cwd(), 'package.json'),
-      (error: NodeJS.ErrnoException | null, buffer: Buffer) => {
-        if (error !== undefined && error !== null) {
-          reject(error);
-        } else {
-          resolve(JSON.parse(buffer.toString()).dependencies);
+  async displayNestInformation(): Promise<void> {
+    this.displayCliVersion();
+    console.info(chalk.green('[Nest Platform Information]'));
+    try {
+      const dependencies: PackageJsonDependencies = await this.readProjectLockDependencies();
+      this.displayNestVersions(dependencies);
+    } catch {
+      await this.displayNestInformationFromPackage();
+    }
+  }
+
+  async displayNestInformationFromPackage(): Promise<void> {
+    try {
+      const dependencies: PackageJsonDependencies =await this.readProjectPackageDependencies();
+      this.displayNestVersions(dependencies);
+    } catch {
+      console.error(
+        chalk.red(MESSAGES.NEST_INFORMATION_PACKAGE_MANAGER_FAILED),
+      );
+    }
+  }
+
+  displayCliVersion(): void {
+    console.info(chalk.green('[Nest CLI]'));
+    console.info(
+      'Nest CLI Version :',
+      chalk.blue(require('../package.json').version),
+      '\n',
+    );
+  }
+
+  async readProjectLockDependencies(): Promise<PackageJsonDependencies> {
+    return new Promise<PackageJsonDependencies>((resolve, reject) => {
+      readFile(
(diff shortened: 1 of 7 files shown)


jmcdo29:
I just realized as well the error message still references `package.json` so I'll need to update that before the PR is merged.

dkantereivin:
> Kind of depends on your definition of breaking. Though it should be noted that this PR doesn't support PNPM as we need to merge #788 for that. Because of this, PNPM users will lose access to `nest info` until that can be taken up, **or** we have the old functionality still in place (for pnpm and for not finding a lock file).

Regardless of #788, keeping the legacy functionality is a fallback for unrecognized lockfiles seems advantageous. As Kamil mentions in #788, `nest-cli` is theoretically package-manager-agnostic in that _any_ package manager can be used (by manually deleting `node_modules` and lockfile and installing packages).

This also means that if nobody were to add pnpm lockfiles, then `pnpm` users would at least get some `nest info` functionality. This reduces the maintenance cost of pnpm. That being said, I get the feeling that you would be willing to add this functionality (and if not, I would).

actions/info.action.ts
```diff
@@ -77,7 +86,7 @@ export class InfoAction extends AbstractAction {
     );
   }
 
-  async readProjectPackageJsonDependencies(): Promise<PackageJsonDependencies> {
+  async readProjectLockDependencies(): Promise<PackageJsonDependencies> {
```
dkantereivin, line 89: Insignificant, but I think this would be more clear as `readProjectLockfileDependencies`.

jmcdo29: Probably, but I was also trying to keep the name from getting too long.

dkantereivin: I agree, perhaps it's worth including this as a comment? Just because, a "project lock" doesn't exactly refer to anything at all.

jmcdo29: I think the fact that 3 lines later you see `package-lock.json` or `yarn.lock` makes the intention clear. Though I guess in a stack trace it could be _somewhat_ problematic, though a comment wouldn't help that either

dkantereivin: What if you removed the word `project`? It's probably the first assumption once you see "lockfile," although I suppose it doesn't _have_ to be a project.

kamilmysliwiec:
Thanks @jmcdo29! Can we fallback to `package.json` (and information about the dependencies from this file) if **any error is thrown**? In this case, we shouldn't introduce any breaking change for anyone who is using a different package manager than npm or yarn. Also, how complicated would it be to integrate pnpm as well?

andreialecu:
It would probably be much simpler and less problematic to use `require("@nestjs/cli/package.json").version` to get the right version instead of trying to read lockfiles.

jmcdo29:
> Thanks @jmcdo29! Can we fallback to `package.json` (and information about the dependencies from this file) if **any error is thrown**? In this case, we shouldn't introduce any breaking change for anyone who is using a different package manager than npm or yarn. 

That's the current functionality. If there's an error that happens while reading the lockfile, the `package.json` is used instead

> Also, how complicated would it be to integrate pnpm as well?

I'd have to look into the `pnpm-lock.yaml` file and see how to parse that. I can say at the moment that I've tested the current code against a pnpm nest project and it does currently read the `package.json`

jmcdo29:
> It would probably be much simpler and less problematic to use `require("@nestjs/cli/package.json").version` to get the right version instead of trying to read lockfiles. And it will work even without having a lockfile, or with any other package manager.
> 
> Maybe `require.resolve` with the `paths` option if it needs to be resolved from a different folder, then either read `package.json` with `require` or with `readFileSync` and `JSON.parse` (safer)

Hmm, not a bad idea actually. We'd need to keep a running list of all Nest package in the CLI repo. Or we could read the dependencies and devDependencies of the `package.json` and then find all `@nestjs/` dependencies, iterate over them and use the `require(package).version` as you mentioned.

andreialecu:
Just a note, it is theoretically better to use `fs.readFile` and `JSON.parse` than to `require` `package.json`, since in theory `package.json.js` might be executable, and we wouldn't want to execute that. It shouldn't happen unless some sort of malware does it, but still, better to read and parse.

jmcdo29:
Unfortunately had to force push due to some rebasing problems, but got the update done which uses `fs.readFileSync` instead of `require`. Also, there's no more lock file dependency so this works well with any package manager (at least the big three had no problem). Let me know if you have any other concerns

jmcdo29:
@kamilmysliwiec Any way we can make this a part of the v8 push? Just to have better package version reporting by the `nest info` command?

kamilmysliwiec:
LGTM 🙌 Apologies for the delay

Issue #954, opened 2020-11-19
load collection path from `nest-cli.json`


taozhiyuzhuo:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

collection default value is `@nestjs/schematics`. if we want to use custom schematics. I must execute  `nest g -c custom/schematics`

## Expected behavior

- nest cli can load custom schematics path from `nest-cli.json`. if I set the `  "collection": "@custom/schematics"`
when I execute `nest g ` it can load my custom schematics
- it should be better if it can merge my custom schematics to `@nestjs/schematics`.

## What is the motivation / use case for changing the behavior?

Generate files by custom schematics more convenient

kamilmysliwiec:
> nest cli can load custom schematics path from nest-cli.json. if I set the "collection": "@custom/schematics"
when I execute nest g it can load my custom schematics

Setting `"collection": "@custom/schematics"`. Does it not work for you?

taozhiyuzhuo:
> > nest cli can load custom schematics path from nest-cli.json. if I set the "collection": "@custom/schematics"
> > when I execute nest g it can load my custom schematics
> 
> Setting `"collection": "@custom/schematics"`. Does it not work for you?

yes, it does not work. always use `@nestjs/schematics`

kamilmysliwiec:
Would you like to create a PR for this issue?

taozhiyuzhuo:
> Would you like to create a PR for this issue?

sorry for the response you late.

rthomps7:
@kamilmysliwiec 

I don't think this PR is really what you need because t makes it so the `nest-cli.json` configuration will always override the CLI. What we realy need is the `Collection.NESTJS` fallback [here](https://github.com/nestjs/nest-cli/blob/57be6edc6948732fbfe0e7b31e77cea910b90215/commands/generate.command.ts#L61) to pull from the nest-cli.json

kamilmysliwiec:
@rthomps7 would you like to create a PR with a fix?

ledniy:
Hi, @kamilmysliwie could you review [PR](#1166) that addressing that issue? Generally, I've used @rthomps7 suggestion and it worked.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest-cli/pull/1166

PR #956, opened 2020-11-19
#954 load configuration collection path priority

taozhiyuzhuo:

Issue Number: 954

## What is the new behavior?
load`nestjs-cil.json` configuration collection path priority

### DIFF
--- a/actions/generate.action.ts
+++ b/actions/generate.action.ts
@@ -35,7 +35,7 @@ const generateFiles = async (inputs: Input[]) => {
   const spec = inputs.find((option) => option.name === 'spec');
 
   const collection: AbstractCollection = CollectionFactory.create(
-    collectionOption || configuration.collection,
+    configuration.collection||collectionOption ,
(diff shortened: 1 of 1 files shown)


actions/generate.action.ts
```diff
@@ -35,7 +35,7 @@ const generateFiles = async (inputs: Input[]) => {
   const spec = inputs.find((option) => option.name === 'spec');
 
   const collection: AbstractCollection = CollectionFactory.create(
-    collectionOption || configuration.collection,
+    configuration.collection||collectionOption ,
```
jmcdo29, line 38: What would happen if there is a collection option passed via CLI, but collection is configured in the `nest-cli.json`? Wouldn't this always take the file config over the CLI config?

taozhiyuzhuo: for now, the collection option always been `@nestjs/schematics` cause it has the default value. file config does not work.

Issue #992, opened 2020-12-11
Move webpack to peerDependency for Yarn v2 + hot reload


brandon-leapyear:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
Using `@nestjs/cli` with Yarn v2 + hot reload doesn't work

## Expected behavior
Should work

## Minimal reproduction of the problem with instructions
1. `nest new`
1. `yarn set version berry`
1. Add hot reload per docs, except using `run-script-webpack-plugin` instead of `start-server-webpack-plugin`
1. Make change per https://github.com/nestjs/docs.nestjs.com/pull/1568
1. `yarn start:dev`

## What is the motivation / use case for changing the behavior?
Trying to get hot-reload working with Yarn v2, I initially got an error when requiring `webpack` in `webpack-hmr.config.js` because my `package.json` doesn't explicitly list `webpack`. After installing `webpack`, I got a weird error
```
The 'compilation' argument must be an instance of Compilation
```
And after lots of debugging, I realized that this happens because the webpack I installed was a different version (5.10.0) than the one nest-cli depends on (5.9.0).

The solution to fix both of these problems:
1. The Hot-Reload docs should mention to install `webpack` as well
2. `@nestjs/cli` should move `webpack` to its `peerDependencies`

The current workaround is to make sure to use the same version of webpack as nest-cli; i.e. `yarn why webpack` should show the same version for both your workspace and `@nestjs/cli`

## Environment

```

```

kamilmysliwiec:
Would you like to create a PR for this issue?

NearString:
Same here !

brandonchinn178:
@kamilmysliwiec added PRs:
* https://github.com/nestjs/nest-cli/pull/1000
* https://github.com/nestjs/docs.nestjs.com/pull/1621

merceyz:
You can also use packageExtensions to force it into a peer dependency
```yml
packageExtensions:
  "@nestjs/cli@*":
    peerDependencies:
      webpack: '*'
```

brandon-leapyear:
Ah yes. Well regardless, this would fix

PR #1000, opened 2020-12-26
Move webpack to peerDependencies

brandonchinn178:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features) -- https://github.com/nestjs/docs.nestjs.com/pull/1621

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/nest-cli/issues/992

## What is the new behavior?
`nest` commands will now use the user's installed `webpack` dependency

```
[ ] Yes
[x] No
```

## Other information

I've tested this manually with the following repro steps:
1. Clone my fork, checkout `webpack-to-peer-dep`, run `npm install` and `npm run build`
2. In a new folder: `npm install /path/to/nest-cli-fork`
3. `node_modules/.bin/nest new` > use `yarn`
4. `cd` into your new app
5. Edit `package.json` to change `@nestjs/cli` dep to `file:/path/to/nest-cli-fork`
6. `yarn set version berry` and `yarn install`
7. `yarn add -D webpack-pnp-externals run-script-webpack-plugin webpack`
8. Add `webpack-hmr.config.js`:
    ```js
    const webpack = require('webpack');
    const { RunScriptWebpackPlugin } = require('run-script-webpack-plugin');
    const { WebpackPnpExternals } = require('webpack-pnp-externals');

      return {
        ...options,
        entry: ['webpack/hot/poll?100', options.entry],
        watch: true,
        externals: [
          WebpackPnpExternals({ exclude: ['webpack/hot/poll?100'] }),
        ],
        plugins: [
          ...options.plugins,
          new webpack.HotModuleReplacementPlugin(),
          new webpack.WatchIgnorePlugin({
            paths: [/\.js$/, /\.d\.ts$/]
          }),
          new RunScriptWebpackPlugin({ name: options.output.filename }),
        ],
      };
    };
    ```
9. Edit `main.ts` and `start:dev` as usual
10. `yarn start:dev`
11. Edit `app.controller.ts` and see hot reload work

Will update the docs repo shortly


kamilmysliwiec:
What's the reason for making `webpack` a peer dependency? From the very beginning, `webpack` is (and should be) a required dependency. Is there anything we can do to fix this issue without making it a peer dep and introducing a breaking change in the package?

brandonchinn178:
ah maybe I should revert the optional flag then.

A peer dependency is a dependency where your library needs to use the same
installation as the user. In this case, without webpack as a peer dep,
there would be two webpacks installed: we'll call them webpackA for the one
nest installs and webpackB for the one the user installs. This breaks (as
described in the issue) when using hot reload with Yarn 2 because Yarn 2 is
more strict about dependencies. What's happening is nest uses webpackA to
run the config, which the user configures with plugins from webpackB.
webpackA inspects the plugins and finds that they aren't from webpackA (it
checks instanceof webpackA.Compilation, when the plugins are actually
instances of webpackB.Compilation) and fails.

You might think, "This is only an issue for people using Yarn 2 and hot
reload. Why make such a breaking change for a small fraction of the
usecases?" It's a very poor + frustrating experience for the quickstart
instructions to not work, whatever your developer tooling is. Short of
requiring people use npm if they want to use NestJS, you can't ignore this
issue. One possible solution is to just change the docs to explicitly
specify the webpack version in the npm install step to match the webpack
version in nestjs/cli's dependencies... meaning the docs would need to be
updated everytime the webpack dep is upgraded. Not great.

But more broadly, **keeping webpack as a dependency and not a peer dep
relies on incorrect npm behavior**. The nest cli is using webpack as a peer
dep, by virtue of allowing users to plug in their own webpack config. The
other alternative is to remove the `--webpackPath` flag and prevent users
from specifying their own config. But that would require another approach
for hot reload and, back to your original complaint, be an even bigger
breaking change than this. Really, this PR isn't that much of a breaking
change. People using the webpack switch in the nest cli should be
installing webpack anyway.

Also, as a contributor, it wouldve been nicer to have this discussion in
the issue before asking for a PR and me putting in the work. For future
reference

kamilmysliwiec:
By merging this, we would fix potential issues that may arise from using Hot Reload with Yarn Berry but at the same time, introduce a breaking change for any other, existing application and also, break all the existing tutorials/guides/videos in which authors didn't install `webpack` explicitly, correct(?). I'm not sure what we can do here as this is a pretty rough situation in this case.

brandonchinn178:
Do you have a use-case where someone would need webpack installed by nest
cli but they dont have it installed already? My understanding is running
nest without `--webpack` uses some other build system, right? When would
someone use --webpack without using their own webpack config / installing
webpack themselves?

brandonchinn178:
Also, this is not a "potential" issue. It's failing right now, per the
repro steps in the ticket.

kamilmysliwiec:
> When would
someone use --webpack without using their own webpack config / installing
webpack themselves?

In most cases, you shouldn't be installing `webpack` by yourself. It's needed only for more complicated, custom configurations + HMR which is opt-in as well. For ~90% (ballpark estimate), using the default config should be sufficient.

brandonchinn178:
1. Am I correct in saying `nest build` doesn't use webpack?
2. Is it correct to say that webpack is only used by `nest build --webpack`?
3. When would one use `nest build --webpack` without using their own
config? (and using the default config you mentioned)

kamilmysliwiec:
> 1. Am I correct in saying `nest build` doesn't use webpack?

This depends on the CLI mode you have enabled. In the monorepo mode, `nest build` uses `webpack` by default.

> 2. Is it correct to say that webpack is only used by `nest build --webpack`?

Nope

> 3. When would one use `nest build --webpack` without using their own
config? (and using the default config you mentioned)

- in monorepo mode (webpack enabled by default)
- in a single app mode: to bundle the application and have a single, executable JS file
- in a single app mode + multiple libs (generated with `nest g lib`) to properly compile & bundle app with libs

brandonchinn178:
I see. Thanks for clarifying.

My alternate recommendation, then, would be to remove the webpackPath
functionality and update the hot reload docs (are there any other docs that
use webpackPath?) to only support the "Without CLI" path. This would still
be a breaking change, so it's a question of which breaking change you want
(ostensibly, whichever one you view as affecting fewer people).

brandonchinn178:
@kamilmysliwiec what do you think about my latest suggestion to update the hot-reload docs to only support the "Without CLI" path?

We don't have to remove `--webpackPath` yet, but I would recommend removing it at some point, because it relies on incorrect behavior to work.

kamilmysliwiec:
For those who want to use webpack HMR with CLI, we just updated the Nest CLI and in the latest version, the webpack configuration factory passes the underlying webpack package reference as a second argument. That means you can use it instead of installing `webpack` in your project (and so referencing a different copy). Example here https://docs.nestjs.com/recipes/hot-reload

PR #1223, opened 2021-07-01
feat(): add webpackSourceMap option to build and start commands

jtsshieh:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Right now, only the `--debug` option on the `start` command allows the generation of source maps by webpack.

Issue Number: N/A

## What is the new behavior?
A `--webpackSourceMap` option has been added to the `build` and `start` commands. This allows the generation of webpack source maps based on the user provided type. In addition, the `debug` option on `nest start` will continue to generate `inline-source-map`.

```
[ ] Yes
[X] No
```

### DIFF
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -46,6 +46,9 @@ export class BuildAction extends AbstractAction {
   public async handle(inputs: Input[], options: Input[]) {
     try {
       const watchModeOption = options.find((option) => option.name === 'watch');
+      const webpackSourceMapOption = options.find(
+        (option) => option.name === 'webpackSourceMap',
+      );
       const watchMode = !!(watchModeOption && watchModeOption.value);
 
       const watchAssetsModeOption = options.find(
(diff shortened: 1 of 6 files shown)


kamilmysliwiec:
> Right now, only the --debug option on the start command allows the generation of source maps by webpack.

https://docs.nestjs.com/cli/monorepo#webpack-options

image

You can easily override the default webpack configuration. For example, to set the `devtool` configuration, you could create the following `webpack.config.js` file in the root directory:

```javascript
  return {
    ...options,
    devtool: 'inline-source-map'
  };
};
```

Thanks for your contribution anyway!

jtsshieh:
This solution was actually suggested by jmcdo29 in the discord server. Another possible solution that I was thinking would just add a `debug` flag to build. That would allow `build` to also generate source maps along with `start`, which was my original goal. Could that be implemented instead?

kamilmysliwiec:
There's no reason to add more flags to the CLI commands for things that can be easily configured at the `webpack.config.js` file level.

PR #1434, opened 2021-12-01
refactor: clarify about schematics available on generate command

### LINKED ISSUE #1433: `generate --help` ignores custom collections

### Current behavior

`nest generate --help` always outputs this:

image

(a list of available schematics) even when the schematics from `@nestjs/schematics` aren't available. For instance, using a custom collection

```
$  cat nest-cli.json 
{
  "collection": "../my-collection-that-doesnt-extends-nestjs-schematics",
  "sourceRoot": "src"
}
```

then we can't run, for instance, `nest generate application`

image

---

basically this is due to the following:

https://github.com/nestjs/nest-cli/blob/35df7f8ad339a07364ae6ac0ef720ff90a16c383/commands/generate.command.ts#L77-L83

https://github.com/nestjs/nest-cli/blob/35df7f8ad339a07364ae6ac0ef720ff90a16c383/commands/generate.command.ts#L100

https://github.com/nestjs/nest-cli/blob/35df7f8ad339a07364ae6ac0ef720ff90a16c383/lib/schematics/nest.collection.ts#L124-L128

which doesn't take in count the `collection` option from `nest-cli.json` config file

### Expected behavior

`nest generate --help` should list the schematics available in the loaded collection

But I see three problems here:

- can we fetch all schematics available of a collection programmatically?
- the config file must be parsed every time someone uses `--help`, to retrieve the collection name. Which I think will degraded DX (without thinking of caching things around)
- I don't think we can load the value of `--collection` option on `GenerateCommand#buildDescription` method

So, looks like listing the right set of schematics is not possible.

Then what we could do is indicate that the table lists schematics from the collection `@nestjs/schematics`, not just "Available schematics". This should avoid misusages of NestJS's CLI. Maybe _"Schematics available on @nestjs/schematics collection"_ will be enough?

### Package version

8.1.5

### NestJS version

8.2.3

micalevisk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Closes #1433

## What is the new behavior?

a minor change that could prevent misusages

image

```
[ ] Yes
[x] No
```

### DIFF
--- a/commands/generate.command.ts
+++ b/commands/generate.command.ts
@@ -77,7 +77,9 @@ export class GenerateCommand extends AbstractCommand {
   private buildDescription(): string {
     return (
       'Generate a Nest element.\n' +
-      '  Available schematics:\n' +
+      `  Schematics available on ${chalk.bold(
+        '@nestjs/schematics',
+      )} collection:\n` +
       this.buildSchematicsListAsTable()
     );
   }


kamilmysliwiec:
lgtm

Issue #489, opened 2019-11-30
Fail on schematics when directories starting with "_" gets changed to "-"


ruslanguns:
## I'm submitting a...

<pre><code>
[ ] Regression 
[X] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

When you generate something with schematics, directory's name starting with "_" on result it ignore it and creates with "-".

## Expected behavior

It should respect the "_" or any special character on directory's name.

## Minimal reproduction of the problem with instructions

Just generate something e.g.: 
`nest g co auth/_users/user.module`

![image](https://user-images.githubusercontent.com/19806968/69903522-5296f800-1392-11ea-9d31-03a763ea1f35.png)

## Environment

![image](https://user-images.githubusercontent.com/19806968/69903528-62164100-1392-11ea-8d63-97b8e21b596a.png)

kamilmysliwiec:
Would you like to create a PR for that :)?

ruslanguns:
I feel motivated, will try to study this code as soon as it is possible. 👍 I have an opportunity to leave my name in this incredible project.

dario-rodriguez:
Hello, 
As a hint, it's probably related to this line: https://github.com/nestjs/schematics/blob/2c732fb2718d7b33adfe64415d315aceee497c44/src/lib/controller/controller.factory.ts#L49

It's in the controller.factory but I think this line is present in most schematics.

egodsk:
Hi,
Took a quick look at it. What @dario-rodriguez says might be true once the command is sent to the schematics handler, however the nest-cli also "dasherize" the `name` option, as seen here: https://github.com/nestjs/nest-cli/blob/b4a0ce0356099dbfde4b01fee246e0cda4cbed72/lib/schematics/schematic.option.ts#L25

So that the `name` options is already from the nest-cli sent to the schematic handler with `--name=auth/-users/users` instead of the expected `--name=auth/_users/users` for the following command: `nest g controller auth/_users/users --no-spec`

However while looking at this, my initial thoughts, is that the dasherize command is there for converting camelCase, spaces, and underscores of a the file name to use dashes instead. I have quick couple of questions:
1. Should the behavior be changed such that *both* filenames and directory names are kept in their original format (besides space)? Or should it only be directories, which should not be dasherized?
[...]

PR #1551, opened 2022-03-01
fix(options): allow underscore in names

stanimirovv:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently underscores are turned into dashes.

Issue Number: https://github.com/nestjs/nest-cli/issues/489

## What is the new behavior?
Underscores are supported from nest-cli.
The commands built are with underscores, however, this will not fix the issue entirely since it must also be changed here:
https://github.com/nestjs/schematics/blob/2c732fb2718d7b33adfe64415d315aceee497c44/src/lib/controller/controller.factory.ts#L49

If the PR is accepted I will also make the PR in the other repo.

```
[ ] Yes
[X] No
```

### DIFF
--- a/lib/package-managers/abstract.package-manager.ts
+++ b/lib/package-managers/abstract.package-manager.ts
@@ -4,7 +4,7 @@ import * as ora from 'ora';
 import { join } from 'path';
 import { AbstractRunner } from '../runners/abstract.runner';
 import { MESSAGES } from '../ui';
-import { dasherize } from '../utils/formatting';
+import { normalizeToKebabOrSnakeCase } from '../utils/formatting';
 import { PackageManagerCommands } from './package-manager-commands';
 import { ProjectDependency } from './project.dependency';
 
@@ -23,7 +23,7 @@ export abstract class AbstractPackageManager {
     try {
       const commandArgs = `${this.cli.install} ${this.cli.silentFlag}`;
       const collect = true;
-      const dasherizedDirectory = dasherize(directory);
+      const dasherizedDirectory = normalizeToKebabOrSnakeCase(directory);
       await this.runner.run(
         commandArgs,
         collect,
--- a/lib/package-managers/abstract.package-manager.ts
+++ b/lib/package-managers/abstract.package-manager.ts
@@ -23,11 +23,11 @@ export abstract class AbstractPackageManager {
     try {
       const commandArgs = `${this.cli.install} ${this.cli.silentFlag}`;
       const collect = true;
-      const dasherizedDirectory = normalizeToKebabOrSnakeCase(directory);
+      const normalizedDirectory = normalizeToKebabOrSnakeCase(directory);
       await this.runner.run(
         commandArgs,
         collect,
-        join(process.cwd(), dasherizedDirectory),
+        join(process.cwd(), normalizedDirectory),
       );
       spinner.succeed();
       console.info();
--- a/lib/package-managers/abstract.package-manager.ts
+++ b/lib/package-managers/abstract.package-manager.ts
@@ -1,10 +1,10 @@
-import { dasherize } from '@angular-devkit/core/src/utils/strings';
 import * as chalk from 'chalk';
 import { readFile } from 'fs';
 import * as ora from 'ora';
 import { join } from 'path';
 import { AbstractRunner } from '../runners/abstract.runner';
 import { MESSAGES } from '../ui';
+import { dasherize } from '../utils/formatting';
 import { PackageManagerCommands } from './package-manager-commands';
 import { ProjectDependency } from './project.dependency';
 
--- a/lib/schematics/schematic.option.ts
+++ b/lib/schematics/schematic.option.ts
@@ -1,4 +1,4 @@
-import { dasherize } from '../utils/formatting';
+import { normalizeToKebabOrSnakeCase } from '../utils/formatting';
 
 export class SchematicOption {
   constructor(private name: string, private value: boolean | string) {}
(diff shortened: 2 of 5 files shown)


lib/schematics/schematic.option.ts
```diff
@@ -31,4 +28,13 @@ export class SchematicOption {
         return `${content}${char}`;
       }, '');
   }
+
+  private dasherize(str: string) {
+    const STRING_DASHERIZE_REGEXP = /[\s]/g;
```
micalevisk, line 33:
```suggestion
    const STRING_DASHERIZE_REGEXP = /\s/g;
```

micalevisk:
you can take this one to fix the `new` command as well :smile: Currently, `nest new my_app` creates the `my-app` directory

https://github.com/nestjs/nest-cli/blob/73ba62898e7964d412ef58c3640bbb4307f54afe/actions/new.action.ts#L77

https://github.com/nestjs/nest-cli/blob/73ba62898e7964d412ef58c3640bbb4307f54afe/lib/package-managers/abstract.package-manager.ts#L26

in this case, you'll need to extract the `dasherize` method. Maybe to a new file called `lib/utils/formatting.ts`

micalevisk:
@stanimirovv if you don't want to fix that one it's fine. Let us know.

stanimirovv:
@micalevisk agreed, I will ping you when I've pushed the changes.

stanimirovv:
@micalevisk I've applied the changes.

The fix still doesn't work from e2e point of view since it also needs to be adjusted in the schematics repo as well: https://github.com/nestjs/schematics/blob/2c732fb2718d7b33adfe64415d315aceee497c44/src/lib/controller/controller.factory.ts#L49

Once we have wrapped this PR I will move on to it.

micalevisk:
indeed, I just notice this. That wrong behavior on `nest new my-app` is due to this line: https://github.com/nestjs/schematics/blob/2c732fb2718d7b33adfe64415d315aceee497c44/src/lib/application/application.factory.ts#L21 as the `dasherize` utility from `@angular-devkit/core` will turn `my_app` into `my-app` (and this is [the expected behavior](https://github.com/angular/angular-cli/blob/c4d3dfa14f5d145b831857645f27c056c893526a/packages/angular_devkit/core/src/utils/strings.ts#L38))

So the name `dasherize` of the function you wrote isn't good but naming is hard :smiling_face_with_tear:

lib/utils/formatting.ts
```diff
@@ -0,0 +1,8 @@
+export function dasherize(str: string) {
```
micalevisk, line 1:
```suggestion
export function keepKebabOrSnakeCase(str: string) {
```

maybe this name describes better?

---

```js
// kebab
dasherize('a-b-c') === 'a-b-c',
// snake
dasherize('a_b_c') === 'a_b_c',

dasherize('a b c') === 'a-b-c',
dasherize('a-b c') === 'a-b-c',
dasherize('a_b c') === 'a_b-c',
dasherize('a-b_c') === 'a-b_c',
```

stanimirovv: I like the name, but the function transforms, right , I feel this needs to be showed in the name.

if you have camelCase it is transformed... hm... maybe something generic like transformCamelCase ? or transformName ? 

Honestly I don't really like my suggestions, but we deffinetly need to adjust the function name.

micalevisk: I guess you could choose any other than just `dasherize`, and write JSDoc comment to describe its for now

```js
transformToKebabOrSnakeCase('fooBar') === 'foo-bar' // camel
transformToKebabOrSnakeCase('FooBar') === 'foo-bar' // pascal
transformToKebabOrSnakeCase('foo bar') === 'foo-bar' // spaced

// or maybe 'normalizeToKebabOrSnakeCase'
```

stanimirovv: I really liked including 'normalize' in the name, nice suggestion.

I've updated the code. I am not really confident in the jsdoc description, but we can always adjust it.

lib/utils/formatting.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ *
+ * @param str
+ * @returns formated string
+ * @description normalizes input to supported path and file name format.
+ * Changes camelCase strings to kebab-case, replaces spaces with dash and keeps underscores.
+ */
+
```
micalevisk, line 8:
```suggestion
```

just a minor formatting fix

lib/package-managers/abstract.package-manager.ts
```diff
@@ -23,7 +23,7 @@ export abstract class AbstractPackageManager {
     try {
       const commandArgs = `${this.cli.install} ${this.cli.silentFlag}`;
       const collect = true;
-      const dasherizedDirectory = dasherize(directory);
+      const dasherizedDirectory = normalizeToKebabOrSnakeCase(directory);
```
micalevisk, line 26:
```suggestion
      const normalizedDirectory = normalizeToKebabOrSnakeCase(directory);
```

(and then rename the occurrences below)

stanimirovv: Nice find, I checked the other places - this was the only variable rename required.
I've applied a fix and pushed it.

micalevisk (approved):

stanimirovv:
@micalevisk no action is required on my part as of now, right ?

micalevisk:
@stanimirovv right :) thank you. You can start the fix on https://github.com/nestjs/schematics side

kamilmysliwiec:
LGTM

Issue #1564, opened 2022-03-11
Inconsistent status codes returned on errors from nest build


schmoesta:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

In some cases the `nest build` command will still exit with code 0, even if it encountered an error.

Possibly related: #423, #1044

### Minimum reproduction code

https://github.com/nestjs/typescript-starter

### Steps to reproduce

1. `npm i`
2. `rm tsconfig.build.json`
3. `npx nest build; echo $?`

This will result in:
```
 Error  Could not find TypeScript configuration file "tsconfig.build.json". Please, ensure that you are running this command in the appropriate directory (inside Nest workspace).

0
```

On the other hand:
1. `npm i`
2. `rm src/app.module.ts`
3. `npx nest build; echo $?`

results in:
```
src/main.ts:2:27 - error TS2307: Cannot find module './app.module' or its corresponding type declarations.

2 import { AppModule } from './app.module';
                            ~~~~~~~~~~~~~~

Found 1 error(s).

1
```

### Expected behavior

The command should exit with code 1 if there's an error.

### Package version

8.2.2

### NestJS version

8.3.1

### Node.js version

v16.14.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

### Other

I'm not sure about the underlying mechanics here, but I noticed that there are some other cases where the status code is inconsistent:

```
$ npx nest foo; echo $?

 Error  Invalid command: foo
See --help for a list of available commands.

1

$ npx nest g foo; echo $?
Invalid schematic "foo". Please, ensure that "foo" exists in this collection.
0
```

stanimirovv:
Interesting issue. I couldn't reproduce it on ubuntu with bash.

For the "Other" section how many times do you need to run the command to simulate the error ?

micalevisk:
this happens because here:

https://github.com/nestjs/nest-cli/blob/6ce99a06152a12c9eeb483aaae3889fe0d48c327/lib/compiler/helpers/tsconfig-provider.ts#L13

the exit code isn't changed.

Nor here:

https://github.com/nestjs/nest-cli/blob/d7717c9bd69cc5885374e7337fd7eb13008e1701/actions/build.action.ts#L59-L65

schmoesta:
> Interesting issue. I couldn't reproduce it on ubuntu with bash.
> 
> For the "Other" section how many times do you need to run the command to simulate the error ?

[...]

PR #1566, opened 2022-03-13
fix: fail on nest add results in non 0 exit code

stanimirovv:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[X] Bugfix
```

Failure to add:

Failure on `add` doesn't return a non zero exit code
```
 npm run build && node bin/nest.js add asd ; echo $?

> @nestjs/cli@8.2.1 build
> tsc

▹▸▹▹▹ Package installation in progress... ☕
Failed to execute command: npm install --save asd@latest
✖ Package installation in progress... ☕
Unable to install library asd because package did not install. Please check package name.
0
```

Issue Number:
It has to do with https://github.com/nestjs/nest-cli/issues/1564 but does not adress the specific issues it has there.

## What is the new behavior?
add fails with a non zero exit code
```
npm run build && node bin/nest.js add asd ; echo $?

> @nestjs/cli@8.2.1 build
> tsc

▹▸▹▹▹ Package installation in progress... ☕
Failed to execute command: npm install --save asd@latest
✖ Package installation in progress... ☕
Unable to install library asd because package did not install. Please check package name.
1
```

```
[X] No
```

### DIFF
--- a/actions/add.action.ts
+++ b/actions/add.action.ts
@@ -44,7 +44,7 @@ export class AddAction extends AbstractAction {
           MESSAGES.LIBRARY_INSTALLATION_FAILED_BAD_PACKAGE(libraryName),
         ),
       );
-      throw new Error();
+      throw new Error(MESSAGES.LIBRARY_INSTALLATION_FAILED_BAD_PACKAGE(libraryName));
     }
   }
 
--- a/actions/add.action.ts
+++ b/actions/add.action.ts
@@ -3,19 +3,19 @@ import { Input } from '../commands';
 import { getValueOrDefault } from '../lib/compiler/helpers/get-value-or-default';
 import {
   AbstractPackageManager,
-  PackageManagerFactory,
+  PackageManagerFactory
 } from '../lib/package-managers';
 import {
   AbstractCollection,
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
This doesn't seem to be related to the mentioned issue(?)

stanimirovv:
@kamilmysliwiec it is related since it fixes the rest of it instances where there is an issue with the status code.

I will hapily extend the scope and cover the cases described in the issue, I have already have the context.

My question before I do the implementation is:
Is it okay to add a "global" try/catch in the `bootstrap` function and set the status to 1 (this is what I prefer)

OR

should I track down the different `throw` statements to make sure they eventually get to a catch that sets the status to 1 ?

kamilmysliwiec:
I'd start by making sure that `nest build` always exits with the status set to 1 when build fails (as this is the most crucial fix atm) and then we can think if there are any other improvements we should/could make

stanimirovv:
@micalevisk @kamilmysliwiec ready for review

kamilmysliwiec:
LGTM

Issue #1628, opened 2022-04-28
New application with collection already containing a .gitignore fails


maxgaurav:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When setting up a new application using a collection that already contains a **.gitignore** file. The git init process would fail.

![image](https://user-images.githubusercontent.com/10531584/165780732-4e60d738-8fb4-4f38-91f6-214503583c59.png)

### Minimum reproduction code

https://gist.github.com/maxgaurav/c6260d956d5fb7d124f565959406e957

### Steps to reproduce

```bash

npm install -g @nestjs/cli
npm install -g @maxgaurav/nestjs-template-schematics

nest new sample-content
```

### Expected behavior

When a .**gitignore** file already exists it should not try to overwrite with default git content. This actions leads to 

### Package version

8.2.5

### Node.js version

v16.14.2

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

It looks like the action where **.gitignore** is being initialized is not considering the fact a **.gitignore** file can already be there

jmcdo29:
If you're using `nest new` in a directory that already is using `git` you can use the `--skip-git` flag to tell Nest not to worry about the git setup

maxgaurav:
## Findings

The main call to init git passes but when the next which is to create gitignore file fails.

```typescript
    if (!isDryRunEnabled) {
      if (!shouldSkipGit) {
        await initializeGitRepository(projectDirectory); // passes and setups all necessary git init actions
        await createGitIgnoreFile(projectDirectory); // fails as it assumes no .gitignore file would exists
      }

      printCollective();
    }
 ```

Referencing file: https://github.com/nestjs/nest-cli/blob/94b0c037134a59f7bf3fcc1fd2e7b8f14020141f/actions/new.action.ts#L207

The **createGitIgnoreFile**  function does not consider the fact the files copied through schematics may already contain a .gitignore file. I think this is written this way as the default application strucuture in @nestjs/schmatics does not include the .gitignore file. In my collection overide of the schematics I have provided my own gitignore and its causing failure in completion of command.

```typescript
/**
 * Write a file `.gitignore` in the root of the newly created project.
 * `.gitignore` available in `@nestjs/schematics` cannot be published to
 * NPM (needs to be investigated).
 *
 * @param dir Relative path to the project.
 * @param content (optional) Content written in the `.gitignore`.
 *
[...]

PR #1630, opened 2022-04-29
add file exists check to prevent override of existing .gitignore file…

maxgaurav:
… within collection

## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md

## PR Type

```
[ X ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When attempting to create new application through nest cli when collection is provided which contains a gitignore file either fails with git repository init failure or existing gitignore file being overridden by default nestjs template content.
Issue Number: [#1630](https://github.com/nestjs/nest-cli/issues/1628)

## What is the new behavior?
The create git ignore file will skip creation if the file already exists

```
[ ] Yes
[ X ] No
```

### DIFF
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -203,7 +203,12 @@ const initializeGitRepository = async (dir: string) => {
 const createGitIgnoreFile = (dir: string, content?: string) => {
   const fileContent = content || defaultGitIgnore;
   const filePath = join(process.cwd(), dir, '.gitignore');
-  return promisify(fs.writeFile)(filePath, fileContent);
+  return fileExists(filePath).then(exists => {
+    if (!exists) {
+      return promisify(fs.writeFile)(filePath, fileContent);
+    }
+    return;
+  });
 };
 
 const printCollective = () => {
@@ -249,5 +254,15 @@ export const retrieveCols = () => {
     return defaultCols;
   }
 };
+
+const fileExists = (path: string) => {
+  return promisify(fs.access)(path).then(() => true).catch((err: any) => {
+    if (err.code === 'ENOENT') {
+      return false;
+    }
+
+    return Promise.reject(err);
+  });
+}
 
 export const exit = () => process.exit(1);
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -255,7 +255,7 @@ export const retrieveCols = () => {
 };
 
 const fileExists = (path: string) => {
-  try{
+  try {
     fs.accessSync(path);
     return true;
   } catch (err: any) {
@@ -265,6 +265,6 @@ const fileExists = (path: string) => {
 
     throw err;
   }
-}
+};
 
 export const exit = () => process.exit(1);
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -203,12 +203,11 @@ const initializeGitRepository = async (dir: string) => {
 const createGitIgnoreFile = (dir: string, content?: string) => {
   const fileContent = content || defaultGitIgnore;
   const filePath = join(process.cwd(), dir, '.gitignore');
-  return fileExists(filePath).then(exists => {
-    if (!exists) {
-      return promisify(fs.writeFile)(filePath, fileContent);
-    }
+
+  if (fileExists(filePath)) {
     return;
-  });
+  }
+  return promisify(fs.writeFile)(filePath, fileContent);
 };
 
 const printCollective = () => {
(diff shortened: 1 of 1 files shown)


jmcdo29 (approved):

Tony133 (approved):

actions/new.action.ts
```diff
@@ -203,7 +203,12 @@ const initializeGitRepository = async (dir: string) => {
 const createGitIgnoreFile = (dir: string, content?: string) => {
   const fileContent = content || defaultGitIgnore;
   const filePath = join(process.cwd(), dir, '.gitignore');
-  return promisify(fs.writeFile)(filePath, fileContent);
+  return fileExists(filePath).then(exists => {
+    if (!exists) {
+      return promisify(fs.writeFile)(filePath, fileContent);
+    }
+    return;
+  });
```
micalevisk, line 211: why not use async/await here

```suggestion
  const gitIgnoreFileExists = await fileExists(filePath);
  if (!gitIgnoreFileExists) return promisify(fs.writeFile)(filePath, fileContent);
```

actions/new.action.ts
```diff
@@ -203,7 +203,12 @@ const initializeGitRepository = async (dir: string) => {
 const createGitIgnoreFile = (dir: string, content?: string) => {
```
micalevisk, line 203:
```suggestion
const createGitIgnoreFile = async (dir: string, content?: string) => {
```

kamilmysliwiec:
🏓

actions/new.action.ts
maxgaurav: Hi sorry for the dealy in response. Got stuck in my main project. I have no problem doing it this way, I was just trying to get the flow what was already there. If needed I can change.

micalevisk: I believe async/await will be better to follow

actions/new.action.ts
```diff
@@ -250,4 +255,14 @@ export const retrieveCols = () => {
   }
 };
 
+const fileExists = (path: string) => {
+  return promisify(fs.access)(path).then(() => true).catch((err: any) => {
```
micalevisk, line 259: actually, I don't see any reason to use promises here because all the other tasks will need to wait the `createGitIgnoreFile` one

So can you change `fs.access` with [`fs.accessSync`](https://nodejs.org/api/fs.html#fsaccesssyncpath-mode)?

maxgaurav: Yes

maxgaurav: Just changed the flow to use the synchronous access and mainly reverted the createGitIgnoreFile to be mostly the same code except adding the check to file exist then it returns.

micalevisk: can you run `npm ci && npm run format` to fix formatting issues?

micalevisk: but only commit the `actions/new.action.ts` xD

micalevisk: nvm, let me help you with that. I just ran that command myself

actions/new.action.ts
```diff
@@ -250,4 +254,17 @@ export const retrieveCols = () => {
   }
 };
 
+const fileExists = (path: string) => {
+  try{
```
micalevisk, line 258:
```suggestion
  try {
```

actions/new.action.ts
```diff
@@ -250,4 +254,17 @@ export const retrieveCols = () => {
...
+    fs.accessSync(path);
+    return true;
+  } catch (err: any) {
+    if (err.code === 'ENOENT') {
+      return false;
+    }
+
+    throw err;
+  }
+}
```
micalevisk, line 268:
```suggestion
};
```

micalevisk (changes requested):

micalevisk (approved):

actions/new.action.ts
maxgaurav: On a separate note, this whole gitignore problem is there as the npm does not copy .gitignore files to the distributions. Since we are using angular schematics which allows template prefix/suffix to be added which can be automatically removed  through the schematics build process. I have achieved the same for my collection 
Reference: [https://github.com/maxgaurav/nestjs-template-schematics](https://github.com/maxgaurav/nestjs-template-schematics)

Would it be okay that do the same for the nestjs schematics as well by opening a PR which can allow managing of .gitignore file through schematics.

micalevisk: sure! I'll close the PR https://github.com/nestjs/nest-cli/pull/1649 then

Which means that we can close this PR as well, right? then after your fix on that schematics issue, we could open another one to drop the workaround implemented on this repo

maxgaurav: I think this would still be needed as other's who have made their custom collection may be thinking of using the default git ignore created from this cli action. Otherwise it needs to be documented that this behavior will go away and they should bring their own gitignore template through their schematics.

maxgaurav: Yes, makes sense. But form then the responsibility of having .gitingore file will be to the schematics, if not provided then gitingore will not be there in the project structure.

maxgaurav: Once I complete the functionality I will map the pull request here for refernce.

Tony133 (approved):

Issue #1806, opened 2022-10-05
Failed to execute command: pnpm install --reporter=silent


wangrongding:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

1. sudo npm i -g @nestjs/cli
2. nest new nestjs-demo
3. <img width="972" alt="image" src="https://user-images.githubusercontent.com/42437658/194120168-9ce81da8-c680-404e-ab37-785642295baa.png">

### Minimum reproduction code

no

### Steps to reproduce

1. sudo npm i -g @nestjs/cli
2. nest new nestjs-demo
3. <img width="972" alt="image" src="https://user-images.githubusercontent.com/42437658/194120168-9ce81da8-c680-404e-ab37-785642295baa.png">

### Expected behavior

`nest new nestjs-demo` works fine

### Package version

9.1.4

### Node.js version

v16.16.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

micalevisk:
can you run `pnpm install` manually to see what's the error?

jmcdo29:
Most likely due to peer deps. pnpm v7 is strict on them by default. I have `strict-peer-dependencies=false` in a global `.npmrc` to get around it, but we should probably update the flag for pnpm installs

wangrongding:
> can you run `pnpm install` manually to see what's the error?

It only happens when running `nest new xxx`

kamilmysliwiec:
Would you like to create a PR for this issue @jmcdo29?

jmcdo29:
> Would you like to create a PR for this issue @jmcdo29?

You got it! I'll try to get something up soon

Just-imp:
this issue is not sloved
### environment
windows 11 22H2  22621.1413
- node v16.18.1
- nest v9.3.0
- pnpm 8.1.0
### issue 
![image](https://user-images.githubusercontent.com/65442050/229398777-fb8a8691-809d-4fe0-b2a1-69c51b9b8359.png)

Just-imp:
> this issue is not sloved
> 
> ### environment
> windows 11 22H2 22621.1413
> 
> * node v16.18.1
> * nest v9.3.0
> * pnpm 8.1.0
> 
> ### issue
> ![image](https://user-images.githubusercontent.com/65442050/229398777-fb8a8691-809d-4fe0-b2a1-69c51b9b8359.png)

[...]

PR #1820, opened 2022-10-27
fix: update pnpm install flags

jmcdo29:
pnpm v7.0.0 until 7.13.5 uses `--strict-peer-dependencies=true` which causes a failure when not all `peerDependencies` are installed. By setting the flag explicitly, this will no longer be a problem for new users.

fix: #1806

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When a user uses pnpm 7.0.0 to 7.13.5 the install portion of `nest new` fails because of the `strict-peer-dependencies` option being set inherently to `true`, as we do not add every peer dependency to the `package.json`

Issue Number: #1806

## What is the new behavior?

`nest new` will no longer fail for `pnpm` v7.0.0 to 7.13.5. Later versions it should be set to `false` by default anyways

```
[ ] Yes
[X] No
```


kamilmysliwiec:
lgtm

PR #1907, opened 2023-01-29
refactor: collection-factory create method

recepkefelii:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[] Code style update (formatting, local variables)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
The original code used a switch statement to determine which type of AbstractCollection instance to return. The new code uses a conditional if statement to perform the same task, which makes the code more concise and readable. Additionally, the instantiation of the SchematicRunner object was moved outside of the conditional statement, so that it only needs to be created once, improving the performance of the code.

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/schematics/collection.factory.ts
+++ b/lib/schematics/collection.factory.ts
@@ -7,17 +7,12 @@ import { NestCollection } from './nest.collection';
 
 export class CollectionFactory {
   public static create(collection: Collection | string): AbstractCollection {
-    switch (collection) {
-      case Collection.NESTJS:
-        return new NestCollection(
-          RunnerFactory.create(Runner.SCHEMATIC) as SchematicRunner,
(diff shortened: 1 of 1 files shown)


micalevisk:
> so that it only needs to be created once, improving the performance of the code.

the old version creates it once too as the 'case' is evaluated only one time.

But yours is more clean to me.

micalevisk (approved):

recepkefelii:
> > so that it only needs to be created once, improving the performance of the code.
> 
> the old version creates it once too as the 'case' is evaluated only one time.
> 
> But yours is more clean to me.

thanks

PR #2042, opened 2023-04-22
perf(lib)!: prioritize 'nest-cli.json' file over all the others name patterns

micalevisk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

I believe that the most common name for nestjs CLI config file is `nest-cli.json`. Still, it is the last one being looked up when loading the config file over 3 others patterns.

I can't tell if this is the desired behavior as it is not documented.

## What is the new behavior?

The file `nest-cli.json` is now the first one to be selected among all the others names which are:

- `.nestcli.json`
- `nest-cli.json`
- `nest.json`

```
[x] Yes
[ ] No
```

for instance, people might be using a private/local `.nest-cli.json` (or any other config file) along with `nest-cli.json` (which is public) in their projects for whatever reason. We don't want to break them.

Since this was never documented(?), we could think that this isn't a breaking change as well tho.

### DIFF
--- a/lib/configuration/nest-configuration.loader.ts
+++ b/lib/configuration/nest-configuration.loader.ts
@@ -10,9 +10,9 @@ export class NestConfigurationLoader implements ConfigurationLoader {
     const content: string | undefined = name
       ? await this.reader.read(name)
       : await this.reader.readAnyOf([
+          'nest-cli.json',
           '.nestcli.json',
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #2046, opened 2023-04-23
refactor: use native fs promise-based API instead of promisify & fix formatting

micalevisk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

- there are few places where the callback version of native nodejs modules are used
- I found one place in which a string concatenation for file path was used when we could use the cross-platform version

## What is the new behavior?

- replace those calls of `fs.` with their Promise-based API since they were introduced in nodejs v10.0.0
- replace `"${foo}/${bar}"` with `path.join(foo, bar)` for cross-platform compatibility sake
- add unit tests for `PackageManagerFactory.find`
- run `npm run format`

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/package-managers/package-manager.factory.ts
+++ b/lib/package-managers/package-manager.factory.ts
@@ -26,10 +26,14 @@ export class PackageManagerFactory {
       const files = await fs.promises.readdir(process.cwd());
 
       const hasYarnLockFile = files.includes('yarn.lock');
-      if (hasYarnLockFile) return this.create(PackageManager.YARN);
+      if (hasYarnLockFile) {
+        return this.create(PackageManager.YARN);
+      }
 
       const hasPnpmLockFile = files.includes('pnpm-lock.yaml');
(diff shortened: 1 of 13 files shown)


lib/package-managers/package-manager.factory.ts
```diff
@@ -20,22 +20,20 @@ export class PackageManagerFactory {
...
-        }
-      });
-    });
+    const DEFAULT_PACKAGE_MANAGER = PackageManager.NPM;
+
+    try {
+      const files = await fs.promises.readdir(process.cwd());
+
+      const hasYarnLockFile = files.includes('yarn.lock');
+      if (hasYarnLockFile) return this.create(PackageManager.YARN);
```
kamilmysliwiec, line 29:
```suggestion
      if (hasYarnLockFile) {
         return this.create(PackageManager.YARN);
      }
```
etc

kamilmysliwiec:
lgtm

PR #2180, opened 2023-07-16
refactor: replace command inputs and options by a dictionary-like high-level data abstraction

micalevisk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

I notice that all operations on command's inputs and options are using `.find()` to search for the first input/option found

## What is the new behavior?

replace the usage of `Input[]` from `.command.ts` and `.action.ts` files by a new abstraction called `CommandContext` which hides the complexity of registering and retrieving some input/operation in a command lifecycle.

this change is intended to improve:

- code readability  
Instead of having a bunch of calls like `inputs.find((o) => o.name === 'foo')` we now have just one `inputs.resolveInput('foo')`, which is also slightly fast as it uses JS `Map` under the hood. This make it even clear that we shouldn't be define the same input name more than one time. I didn't write a validation for this scenario for [micro]performance sake but we could easily introduce that if needed
- type safety  
Instead of relying on type assertion and non-null assertions, we're using generics to make the code less error prone at compile time. The present solution doesn't differ that much from type assertions but there is a room to improve that easily by now

```
[ ] Yes
[x] No
```


kamilmysliwiec:
I'd love to finally get rid of these confusing Input/Output abstractions - these are leftovers from the first version of the CLI 😄

micalevisk:
in order to try this version better, I've published this branch's version as  
`micalevisk-nestjs-cli` package.

commands/command-storage.ts
```diff
@@ -0,0 +1,71 @@
+export interface CommandStorageEntry<
+  TValue extends boolean | string = boolean | string,
+> {
+  readonly name: string;
+  readonly value: TValue;
+  readonly options?: any;
+}
+
+export class CommandStorage {
```
kamilmysliwiec, line 9: very nitpicky but command storage name makes me think it somewhat holds the "command" reference while this isn't true (it represents the command context/options)

micalevisk: good point

naming is hard, you know haha

I'll think of another name for it

kamilmysliwiec:
This introduces a regression when you run `nest new` without passing the application name as an argument

micalevisk:
I'll see if I can introduce this change in the current version of the main instead

Issue #462, opened 2019-11-13
Add configuration option to generate files with PascalCase instead of the default naming convention


bbenoist:
## I'm submitting a...

- [x] Feature request

## Current behavior

As of now, if I generate a new file with `nest g`, its is named `name.type.ts` (e.g. `my.module.ts`).

I'd like the tool to create a file named `NameType.ts` (e.g. `MyModule.ts`) so that it matches the name of the class it contains.

## Expected behavior

Ideally, a `fileNameCase` configuration option in `nest-cli.json` of type `"dot" | "pascal" | "camel" | "snake" | "kebab"` would be nice.

## Minimal reproduction of the problem with instructions

`nest g mo my` generates a file named `my.module.ts`

## What is the motivation / use case for changing the behavior?

All of my company source files are named with the same name as the classes that they include. It's a common convention we see in a lot of OOP projects.

bbenoist:
Side-demand. It would be nice to also support naming test files with `.test.ts` instead of `.spec.ts` via a `testFileExtension` option too.

kamilmysliwiec:
> Side-demand. It would be nice to also support naming test files with .test.ts instead of .spec.ts via a testFileExtension option too.

We're tracking this here https://github.com/nestjs/schematics/issues/143

egodsk:
If a `fileNameCase` is introduced, should it only affect filenames, as suggested by the option name? What about directory names, if you specify that in the generate command, such as `nest g controller auth-dir/users_dir/users --no-spec`?

Should there also be introduced a `directoryNameCase` option? or should it just be a `nameCase` option that affect both files and directories? 

Something like https://www.npmjs.com/package/change-case could help implement the solution without having to worry about correct case convertions

cyrdax:
this would be a very nice addition.  having to switch to kabob case is a huge pain for our team

donnyroufs:
Is there any news on this?

oonsamyi:
Is there any news on this?(

hydra13:
Is there any news on this? Do you need any help?

kamilmysliwiec:
Contributions are more than welcome! @hydra13

erie0210:
Hi, is this issue still on going? I need this feature (with camelcase) and want to take a look if it's still valid.  @kamilmysliwiec

bennycode:
@kamilmysliwiec can you point me to the code which generates the files? I want to try implementing support for file names in PascalCase. 🙂

[...]

PR #2208, opened 2023-07-31
feat: initial support for case type selection

espoal:
## Summary

This is the second of a multi PR effort aiming at supporting custom name casing conventions, as detailed [here](https://github.com/nestjs/nest-cli/issues/462).

1. The first [PR](https://github.com/nestjs/schematics/pull/1501) attempts to update `@nestjs/schematics` to accept a `caseType` option when generating a new file.
2. In [this](https://github.com/nestjs/nest-cli/pull/2208) PR we add a `caseNaming` field to `GenerateOptions` 
3. [This](https://github.com/nestjs/docs.nestjs.com/pull/2886) is the PR to document the new option
4. Future PRs will extend this feature in term of supported actions, tests and configuration

For now I decided to support only controllers, to reduce the changes size and make reviewing easier. If the approach will be approved I will make sure to extend it to other objects and to add more tests.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently `@nestjs/cli` uses `kebab-or-snake` [case](https://github.com/nestjs/nest-cli/blob/be5aab2940af67886acf6175097b066766632597/lib/utils/formatting.ts#L8), as mentioned in this [issue](https://github.com/nestjs/nest-cli/issues/462).

Issue Number: 462

## What is the new behavior?

The goal is to add a `caseNaming` option to `nest-cli.json` so that generated files will follow the desired name casing convention.

```
[ ] Yes
[x] No
```

### How to run

- Install deps and build:
```
npm install
npm run build
```
- Build updated `@nestjs/schematics`
Clone [this](https://github.com/nestjs/schematics/pull/1501) branch, and do:

```
npm install
npm run build
```
- Install the updated `@nestjs/schematics` package
Copy the `dist` folder in `@nestjs/nest-cli` deps: `nest-cli/node_modules/@nestjs/schematics/dist/`

- Run it with one of those commands:
`node bin/nest.js new test-nest -s --caseNaming=camel` to generate a custom `nest-cli.json`
`node bin/nest.js g co keb-pap` in a folder with the correct `nest-cli.json`


lib/schematics/schematic.option.ts
```diff
@@ -1,13 +1,22 @@
...
+
+const defaultSchematicOptionConfig: SchematicOptionConfig = {
+  caseType: 'kebab-or-snake',
+};
 
 export class SchematicOption {
   constructor(
     private name: string,
     private value: boolean | string,
+    private schematicOptionConfig = defaultSchematicOptionConfig,
```
espoal, line 15: Here I'm adding a `SchematicOptionConfig` to handle a set of generic options we could pass, including `CaseType`.

The first thing I tried was to add something like `SchmaticOption('caseNaming', 'camelCase')`, but I couldn't make it work because when we launch the command `nest generate controller emails --caseNaming=camel` we have two schematic options:
- `SchmaticOption('name', 'emails')`
- `SchmaticOption('suffix', 'controller')`

We need to apply the `camelCase` transform to both, hence why I used a 3 way option. 
I honestly don't like this approach very much, so I would love some feedback.

espoal:
Hey @micalevisk can I help you with anything? I would love a pair programming session to review the pr together and see how to improve it

mwanago:
I think it would make sense to add the `--caseNaming` option to the `new` command that creates a new NestJS application. Would you agree?

micalevisk:
yeah, instead of adding that new option to `generate`, we can move it to `new` and to the configuration file (`nest-cli.json`)

mwanago:
@micalevisk 
Since we usually want to keep the same naming convention throughout the project, storing this option in `nest-cli.json` sounds like a great idea

espoal:
@mwanago Let's split the problem in two parts:

- My main concern is if my approach of decorating `SchemaOption` is a good idea, or if maybe people see a better approach.
- I would totally be ok to use the `new` command. Could you make an example of how a CLI command would be using that? And where the `nest-cli.json` is parsed?

micalevisk:
@espoal 
> I would totally be ok to use the new command. Could you make an example of how a CLI command would be using that?

it would be nothing more than a new flag (`--caseNaming`) for the `new` command that will define the value of some new `caseNaming` (defaults to "kebab") option of the generated `nest-cli.json` (I guess at `GenerateOptions` interface here: https://github.com/nestjs/nest-cli/blob/48939b57814f0353e1759fcabb85fe401b39e793/lib/configuration/configuration.ts#L76C18-L76C33)

Then, when using the `generate` command, that option will be taken into account just like it doe for the `spec` option

espoal:
Sorry for the late reply but I was swamped with my day work.
The PR is far from perfect, but I think I'm slowly getting closer to what you meant @micalevisk .
Some questions:

- I still cannot change the suffix of the file (`KebPap.controller.ts` instead of `KebPapController.ts`). Any advice on how to proceed?
- Now I correctly use the `nest-cli.json` but are we sure we don't want to leave the command line option? Now we have both.

And a slightly bigger topic:
- Now when we use the `new` command the file don't respect the passed convention, because they're taken from a template. The solution IMHO would be to update the `new` command to use a sequence of commands (`configuration`, `generate` ....) instead of the templates. But this seems like a big topic, so maybe you have suggestions.

cc: @mwanago

lib/schematics/schematic.option.ts
espoal: Please resolve this, the approach changed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nestjs/cli",
-  "version": "10.1.11",
+  "version": "10.1.18",
```
micalevisk, line 3: revert this (make sure the lock file is reverted as well)

commands/new.command.ts
```diff
@@ -32,6 +32,10 @@ export class NewCommand extends AbstractCommand {
         Collection.NESTJS,
       )
       .option('--strict', 'Enables strict mode in TypeScript.', false)
+      .option(
+            '--caseNaming [caseType]',
+            `Casing type for generated elements. Available options: "pascal", "camel", "kebab-or-snake" (default).`,
```
micalevisk, line 37: I didn't follow what is the _kebab-or-snake_ type :thinking: 

Also,

```suggestion
            `Casing type for generated files. Available options: "pascal", "camel", "kebab-or-snake" (default).`,
```

as this will cover file names only, as discussed in the issue

micalevisk:
> are we sure we don't want to leave the command line option?

if we can make it work without adding the options to the `generate` command, then yes. Less code to maintain; less options to know about. We can add them in the futurue if needed.

> I cannot see why the pipeline is failing

Does not seems to be releated to your changes. But revert the changes you've made to `package.json` and `package-lock.json`. That's up to Kamil after merging the PR.

> Now when we use the `new` command the file don't respect the passed convention, because they're taken from a template

I didn't follow. The project generated by `new` is too simple. The files will have the same name regardless of the convention because they have only one word. Here I'm talking about source files, not configuration (json) ones.

---


Also, convert this PR to _draft_ until you're done.

commands/new.command.ts
espoal: About the `kebab-or-snake` case:
In the original code there is a case normalization [function](https://github.com/nestjs/nest-cli/blob/master/lib/utils/formatting.ts) that is a mix of kebap and snake case. I kept it for historical purposes, but I also introduced separated kebap and snake normalization functions. I would more than willing to remove it.

espoal:
Thanks Micael for the quick feedback, I really appreciate it. My brain is fried atm from a long week, but I will commit the small changes you suggested and get back on this next week.

espoal:
> if we can make it work without adding the options to the generate command, then yes. Less code to maintain; less options to know about. We can add them in the future if needed.

I think it's not too much code, but I also think that  smaller PRs are better. Let's keep the flag for a future PR.

> The project generated by new is too simple. 

That's a fair point

> Also, convert this PR to draft until you're done.

I still don't think the PR is perfect, but perfection is also the worst enemy of the good, so I would like to ask you @micalevisk  to give the PR a further look. 

Things I would like to do after this review:

- [ ] Decide which case formats we support 
- [ ] Add documentation
- [ ] Add more tests


P.S.: I managed to fix the pipeline :partying_face:

test/lib/utils/formatting.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+type TestSuite = {
+  description: string;
+  input: string;
+  caseType: CaseType;
+  expected: string;
+};
+
+describe('Format strings', () => {
+  const tests: TestSuite[] = [
```
milsosa, line 11: Shouldn't we add test cases for the other casing types?

actions/generate.action.ts
```diff
@@ -44,10 +45,19 @@ const generateFiles = async (inputs: Input[]) => {
   const collection: AbstractCollection = CollectionFactory.create(
     collectionOption || configuration.collection || Collection.NESTJS,
   );
+  
+  const caseType = (
+      configuration?.generateOptions?.caseNaming
+      || 'kebab-or-snake'
```
milsosa, line 51: Should we document that `kebab-or-snake` will be the default option when caseNaming is not specified?

test/lib/schematics/schematic.option.spec.ts
```diff
@@ -38,19 +38,19 @@ describe('Schematic Option', () => {
...
+    },*/
     {
       description: 'should manage classified string option value name',
       option: 'name',
       input: 'MyApp',
       expected: 'my-app',
     },
-    {
+    /*{
       description: 'should manage parenthesis string option value name',
```
espoal, line 54: Atm we don't really respect this. In `package.json` we have the right name `my-(app)`, but at least on linux the folder is actually called `./'my-(app)'` (note the dashes). Same for the `[` character

test/lib/schematics/schematic.option.spec.ts
```diff
@@ -38,19 +38,19 @@ describe('Schematic Option', () => {
       input: 'myApp',
       expected: 'my-app',
     },
-    {
+    /*{
       description: 'should allow underscore string option value name',
```
espoal, line 42: I feel this could be a source of confusion: if I specify the `kebab` case, I would expect the convention to be respected.

espoal:
In the end I decided that the case we should support are: 

- `snake`
- `kebab`
- `camel`
- `pascal`

I would propose to drop the `kebab-or-snake` [case](https://github.com/nestjs/nest-cli/blob/be5aab2940af67886acf6175097b066766632597/lib/utils/formatting.ts#L8) for the reasons outlined [here](issue coming soon).

espoal:
Hey @micalevisk is there anything I could do to speed up the merging of this PR?

micalevisk:
@espoal nop. I'll try to review this ASAP

espoal:
Thanks, I apologies to bother you but I'm really excited about my first nestjs PR :heart:

lib/utils/formatting.ts
```diff
@@ -1,15 +1,50 @@
...
+    case 'snake':
+      return snakeCase(str);
+    case 'kebab-or-snake':
+      return kebabCase(str, { keep: ['_', '@', '/', '.'] })
+    default:
+      throw new Error(`Case type ${caseType} is not supported.`);
+  }
+};
+
+export const formatString = (str: string) => {
```
micalevisk, line 43: can you elaborate here in the code on what this function is supposed to do?

github copilot can help :p

espoal: Dang, I'm surprised copilot understood immediately the code. I'm starting to question reality :D 

Added some JsDoc. Basically this function escapes parenthesis which will then later be removed. 

I see here an opportunity to create `@nestjs/utils` or `@nestjs/stringUtils` because this code is shared across projects, and I feel it would benefit to centralize it so that it's easier to maintain and can be reused by plugin developers. I didn't do it to keep the scope limited, but it shouldn't be too hard to do.

espoal: ok I updated the same code also in `@nestjs/schematics`

RavenHursT:
@espoal are you still going to resolve the conflicts on this PR?

Would love to see this merged and released.  The constant moving and renaming of files and folders is plain tedious..

PR #2237, opened 2023-08-14
fix(swc): should clear watchers if we are not in watch mode, and refact

thomasync:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

If SWC is used as a compiler and the --watchAssets parameter is set, the application never quits after the command has been executed.

When you execute `pnpm run start:cli:swc`, the application never shuts down.
https://github.com/thomasync/swc-nestjs-compilation/tree/clear-watchers-never-called

## What is the new behavior?

Now when you execute `pnpm run start:cli:swc`, the application exits correctly. 

https://github.com/thomasync/swc-nestjs-compilation/tree/clear-watchers-never-called-fixed

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/compiler/swc/swc-compiler.ts
+++ b/lib/compiler/swc/swc-compiler.ts
@@ -69,9 +69,9 @@ export class SwcCompiler extends BaseCompiler {
       await this.runSwc(swcOptions, extras, swcrcFilePath);
       if (onSuccess) {
         onSuccess();
-      } else {
-        extras.assetsManager?.closeWatchers();
       }
+
+      extras.assetsManager?.closeWatchers();
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/compiler/swc/swc-compiler.ts
```diff
@@ -49,29 +49,23 @@ export class SwcCompiler extends BaseCompiler {
       appName,
     );
 
-    if (extras.watch) {
-      if (extras.typeCheck) {
-        this.runTypeChecker(configuration, tsConfigPath, appName, extras);
-      }
-      await this.runSwc(swcOptions, extras, swcrcFilePath);
+    if (extras.typeCheck) {
+      this.runTypeChecker(configuration, tsConfigPath, appName, extras);
```
kamilmysliwiec, line 53: This introduces a breaking change - in watch mode, `runTypeChecker` shouldn't be awaited but when it's turned off, we have to use `await`

thomasync: oh I didn't think it was expected to run swc without waiting for typecheck, which is kind of the expected behavior with typeCheck. I'll re-commit with just the fix.

kamilmysliwiec:
LGTM

PR #2294, opened 2023-09-17
perf: cache loaded configuration file

### LINKED ISSUE #2275: refactor request: avoid useless calls to `NestConfigurationLoader#load` on 'build' (and 'start') commands

while investigating how to reduce the time spent on `build` command just for the cli bootstrapping, I notice that `NestConfigurationLoader#load` method was called twice for the same input:

image

https://github.com/nestjs/nest-cli/blob/802a73c836454c2936fb27b253778020ac9c85f7/lib/configuration/nest-configuration.loader.ts#L9-L37

`name` is always `undefined` for those calls.

somehow refactor the code so that `NestConfigurationLoader#load` will be invoked only once.

that method is using `fs.readFile`+`JSON.parse`. I'd like to avoid this for performance reasons (although minimal, I guess) and to improve the codebase (ie., as a refactor since that seems to be something missed from the legacy version)

micalevisk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: closes #2275 

## What is the new behavior?

now `#load` is being invoked only once instead of twice (`generate.command`, `build.action`) when running `nest build`. And only once instead of three times (`generate.command`, `start.action`, `build.action`) on `nest start`

The downside of this solution is that there's no way to bypass the cache. So a command cannot access a fresh version of the config file unless the CLI process restarts. But AFAIK (and tested) this isn't a feature of the `start --watch`

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/configuration/nest-configuration.loader.ts
+++ b/lib/configuration/nest-configuration.loader.ts
@@ -3,10 +3,24 @@ import { Configuration } from './configuration';
 import { ConfigurationLoader } from './configuration.loader';
 import { defaultConfiguration } from './defaults';
 
+/**
+ * A cache table that maps some reader (by its name along with the config path)
+ * to a loaded configuration.
+ * This was added because several commands relies on the app's config in order
+ * to generate some dynanmic content prior running the command itself.
+ */
+const loadedConfigsCache = new Map<string, Required<Configuration>>();
+
 export class NestConfigurationLoader implements ConfigurationLoader {
   constructor(private readonly reader: Reader) {}
 
   public async load(name?: string): Promise<Required<Configuration>> {
+    const cacheEntryKey = `${this.reader.constructor.name}:${name}`;
+    const cachedConfig = loadedConfigsCache.get(cacheEntryKey);
+    if (cachedConfig) return cachedConfig;
+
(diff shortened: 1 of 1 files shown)


lib/configuration/nest-configuration.loader.ts
```diff
@@ -3,10 +3,24 @@ import { Configuration } from './configuration';
...
+ */
+const loadedConfigsCache = new Map<string, Required<Configuration>>();
+
 export class NestConfigurationLoader implements ConfigurationLoader {
   constructor(private readonly reader: Reader) {}
 
   public async load(name?: string): Promise<Required<Configuration>> {
+    const cacheEntryKey = `${this.reader.constructor.name}:${name}`;
+    const cachedConfig = loadedConfigsCache.get(cacheEntryKey);
+    if (cachedConfig) return cachedConfig;
```
kamilmysliwiec, line 20:
```suggestion
    if (cachedConfig) {
       return cachedConfig;
    }
```

kamilmysliwiec:
LGTM

PR #2325, opened 2023-10-08
refactor: remove dead-code

micalevisk:
- remove the `NestDependencyManager` class as it is not being used
- remove the `isCompilableExtension` compiler helper as it is not being used
- remove the `isError` utility as it is not being used
- remove two unused barrel files
- ran `npm run format`
- a minor fix that doesn't seemed to affect the program anyway: a missing `await`

---

image

image

image


kamilmysliwiec:
LGTM

PR #2446, opened 2024-01-14
fix: exit with an error when some config file was found but lacks on reading permissions

### LINKED ISSUE #2434: odd behavior when the config file can't be read

### Current behavior

When we have some nest config file but the system user running the nodejs process has no permissions to _read_ the file the CLI silently ignores that, which could lead to confusion. For example: someone changed the content of `nest-cli.json` but the CLI ends up using the default values due to missing permissions.

This behavior is due to the following lines:

https://github.com/nestjs/nest-cli/blob/6617460c4f027aa32145af8dcac85ad72920932b/lib/readers/file-system.reader.ts#L21-L25

### Minimum reproduction code

https://github.com/nestjs/typescript-starter

### Steps to reproduce

```bash
git clone https://github.com/nestjs/typescript-starter
cd typescript-starter
npm install

## change the nest-cli.json to an invalid config (eg: change the entryFile)

chmod -r nest-cli.json

npm run start:dev # see no errors
```

### Expected behavior

either exiting with an error due to a non-readable nest config file or a warning message.

Exiting with an error seems better to me but it would be breaking change if the present behavior is a feature not a bug

### Package version

10.2.1

### Other

I can make a PR to address this

micalevisk:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: closes #2434

## What is the new behavior?

exit with an error. See:

https://github.com/nestjs/nest-cli/assets/13461315/2aad7161-08d9-4c56-a2c4-392a04203833

We got the same output for `nest generate` command

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/readers/file-system.reader.ts
+++ b/lib/readers/file-system.reader.ts
@@ -18,8 +18,8 @@ export class FileSystemReader implements Reader {
   ): string | undefined | ReaderFileLackPersmissionsError {
     let firstFilePathFoundButWithInsufficientPermissions: string | undefined;
 
-    for (let idx=0; idx < filenames.length; idx++) {
-      const file = filenames[idx];
+    for (let id = 0; id < filenames.length; id++) {
+      const file = filenames[id];
 
       try {
         return this.read(file);
@@ -35,8 +35,10 @@ export class FileSystemReader implements Reader {
           }
         }
 
-        const isLastFileToLookFor = idx === filenames.length - 1;
-        if (!isLastFileToLookFor) continue;
+        const isLastFileToLookFor = id === filenames.length - 1;
+        if (!isLastFileToLookFor) {
+          continue;
+        }
 
         if (firstFilePathFoundButWithInsufficientPermissions) {
           return new ReaderFileLackPersmissionsError(
--- a/lib/readers/file-system.reader.ts
+++ b/lib/readers/file-system.reader.ts
@@ -5,22 +5,22 @@ import { Reader } from './reader';
 export class FileSystemReader implements Reader {
   constructor(private readonly directory: string) {}
 
-  public list(): Promise<string[]> {
-    return fs.promises.readdir(this.directory);
+  public list(): string[] {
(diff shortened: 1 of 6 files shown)


lib/readers/file-system.reader.ts
```diff
@@ -1,27 +1,52 @@
...
-  public async readAnyOf(filenames: string[]): Promise<string | undefined> {
-    try {
-      for (const file of filenames) {
-        return await this.read(file);
+  public readAnyOf(
+    filenames: string[],
+  ): string | undefined | ReaderFileLackPersmissionsError {
+    let firstFilePathFoundButWithInsufficientPermissions: string | undefined;
+
+    for (let idx=0; idx < filenames.length; idx++) {
```
kamilmysliwiec, line 21:
```suggestion
    for (let idx = 0; idx < filenames.length; idx++) {
```
nit: why idx not just id

lib/readers/file-system.reader.ts
```diff
@@ -1,27 +1,52 @@
...
+        ) {
+          const isInsufficientPermissionsError =
+            readErr.code === 'EACCES' || readErr.code === 'EPERM';
+          if (isInsufficientPermissionsError) {
+            firstFilePathFoundButWithInsufficientPermissions = readErr.path;
+          }
+        }
+
+        const isLastFileToLookFor = idx === filenames.length - 1;
+        if (!isLastFileToLookFor) continue;
```
kamilmysliwiec, line 39:
```suggestion
        if (!isLastFileToLookFor) {
         continue;
        }
```

lib/readers/file-system.reader.ts
micalevisk: I'm used to call _index_ as _idx_ because _i_ is too short, and _id_ looks like 'unique id'

kamilmysliwiec:
lgtm

PR #2495, opened 2024-02-03
feat(new): check for installed package managers

onurravli:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, the CLI tool asks for all package managers (npm, yarn, and pnpm) even if they aren't installed.

## What is the new behavior?

Now, CLI checks for installed package managers and then asks for those that are installed, not for the ones that are not installed.

```
[ ] Yes
[X] No
```

### DIFF
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -64,27 +64,11 @@ export class NewAction extends AbstractAction {
   }
 }
 
-const isNpmInstalled = () => {
+const isThisPackageManagerInstalled = (packageManager: string) => {
   try {
-    execSync('npm -v');
-    return true;
-  } catch (error) {
-    return false;
-  }
-};
-
-const isYarnInstalled = () => {
-  try {
-    execSync('yarn -v');
-    return true;
-  } catch (error) {
-    return false;
-  }
-};
-
-const isPnpmInstalled = () => {
-  try {
-    execSync('pnpm -v');
+    execSync(`${packageManager} -v`, {
+      stdio: 'ignore',
+    });
     return true;
   } catch (error) {
     return false;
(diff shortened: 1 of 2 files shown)


onurravli:
Hey @kamilmysliwiec, can you check this out?

actions/new.action.ts
```diff
@@ -64,6 +64,17 @@ export class NewAction extends AbstractAction {
   }
 }
 
+const isThisPackageManagerInstalled = (packageManager: string) => {
+  try {
+    execSync(`${packageManager} -v`, {
+      stdio: 'ignore',
+    });
+    return true;
+  } catch (error) {
```
micalevisk, line 73:
```suggestion
  } catch {
```

onurravli: 56493f6d.

micalevisk (approved):

Tony133 (changes requested):

actions/new.action.ts
```diff
@@ -163,12 +174,17 @@ const installPackages = async (
 };
 
 const askForPackageManager = async (): Promise<Answers> => {
+  const installedPackageManagers = [];
+  if (isThisPackageManagerInstalled('npm'))
+    installedPackageManagers.push(PackageManager.NPM);
+  if (isThisPackageManagerInstalled('yarn'))
+    installedPackageManagers.push(PackageManager.YARN);
+  if (isThisPackageManagerInstalled('pnpm'))
+    installedPackageManagers.push(PackageManager.PNPM);
```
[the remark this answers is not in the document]
onurravli: c517e57.

Tony133 (approved):

kamilmysliwiec:
Not sure how I feel about this change. Isn't this confusing that some options just won't show up for certain people? Like one might be aware that yarn is a thing but doesn't have it installed on let's say their machine at work. Then they generate a NestJS app and they only see NPM showing up which will make them think using yarn was never an option (simply because it's hidden).

Shouldn't we just leave this decision to the developer since the developer should be aware of what tool they have installed (and even if they don't they can always terminate the command, install it, and start over) and want to use for their projects?

Tony133:
Yes, this is a bit of a "divisive" change, to one 50 percent it can be helpful while the other 50 percent can be confusing 😄

onurravli:
> Not sure how I feel about this change. Isn't this confusing that some options just won't show up for certain people? Like one might be aware that yarn is a thing but doesn't have it installed on let's say their machine at work. Then they generate a NestJS app and they only see NPM showing up which will make them think using yarn was never an option (simply because it's hidden).
> 
> Shouldn't we just leave this decision to the developer since the developer should be aware of what tool they have installed (and even if they don't they can always terminate the command, install it, and start over) and want to use for their projects?

I made this change after I realize there is an error when you continue with yarn even yarn is not installed, as below.

image

If it's not a good idea to hide package managers, it might be a better idea to warn the user if, for example, yarn is not installed, but to build the project (without installing packages).

kamilmysliwiec:
Even if the installation fails, everything else should be properly generated and you should be good to go as long as you manually install dependencies

onurravli:
Yes, I mean the error is not clear. In my firs use of nest-cli, I thought project wasn't built.

kamilmysliwiec:
So maybe instead of hiding package managers we should rather improve our error messages; let users know that even if the installation failed they're generally (almost) all set

kamilmysliwiec:
Feel free to continue in this PR

Issue #1614, opened 2022-04-16
getting `EADDRINUSE` error when using start with watch mode with an async `beforeApplicationShutdown`


micalevisk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

when we have an async `beforeApplicationShutdown` that takes too long to resolve, `nest start --watch` doesn't kill all the processes properly after file changes, which leads to an `EADDRINUSE: address already in use` error

https://user-images.githubusercontent.com/13461315/163654949-53f18d03-fc89-4757-88fe-e1f7d99a8cd1.mp4

### Minimum reproduction code

https://gitlab.com/micalevisk/nestjs-cli-issue-1614

### Steps to reproduce

```bash
git clone https://gitlab.com/micalevisk/nestjs-cli-issue-1614 
cd nestjs-cli-issue-1614 
```

2. `npm ci`
3. `npm run start:dev`
4. change any source file
5. see the error

### Expected behavior

nestjs CLI should kill all processes before restarting the app

### Package version

8.2.x
9.2.x - using `forceCloseConnections: true` doesn't fixes this
10.0.3

### NestJS version

8.4.x
9.3.x
10.0.3

### Node.js version

16.14.2

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

kamilmysliwiec:
Would you like to create a PR to address this issue?

micalevisk:
I don't know how to fix that yet

micalevisk:
I guess this is beyond `nest start --watch`. See here that the app exits right away when we press ctrl+c when starting the app with `nest start` (it works if we use `node`) instead of waiting for `beforeApplicationShutdown` hook:

https://user-images.githubusercontent.com/13461315/164307593-36add876-339a-4498-b412-e34cfcd0604f.mp4

And this is related with #1487 So I believe that if we manage to fix that one, we'll fix this Issue accordingly.

kos984:
I am not sure, but looks like it is related to how nest starts typescript, it just not wait process stop at all, even if you start npm run start you will see same behaviour

TylerOrtiz:
It may not be directly related but I was experiencing this issue (`EADDRINUSE error when using start with watch mode`) constantly while running nest in a container (`nest start --watch`). However it didn't seem to occur while running the project locally (no container). 

Changing the port I was using didn't help but what did seem to help was changing the way I was setting up my WORKDIR and COPY commands. 

[...]

PR #2522, opened 2024-02-21
feat: add flag to disable shell mode for `nest start --watch`

quallrum:
When running on Linux/WSL, `nest start --watch` does not wait the app to shutdown on restart. That leads to two instances of application running simultaneously, especially if it has async shutdown hooks. That commit fixes that adding an option to disable shell mode when spawning child processes so then app will completely shutdown first and only then started again. Thanks @jleverenz for the investigation and solution idea in [#1641](https://github.com/nestjs/nest-cli/issues/1614#issuecomment-1463838659)

fixes nestjs/nest-cli/1614

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Nest does not wait for app to stop when doing hot reload (on files change)

Issue Number: #1614 

## What is the new behavior?

Now running `nest start --watch --noShell` hot reload will first wait for app to gracefully stop

```
[ ] Yes
[x] No
```

### DIFF
--- a/actions/start.action.ts
+++ b/actions/start.action.ts
@@ -72,12 +72,15 @@ export class StartAction extends BuildAction {
         commandOptions,
         defaultConfiguration.sourceRoot,
       );
+      const noShellOption = commandOptions.find((option) => option.name === 'noShell')
+      const useShell = noShellOption ? !noShellOption.value : true;
       const onSuccess = this.createOnSuccessHook(
         entryFile,
         sourceRoot,
         debugFlag,
         outDir,
         binaryToRun,
+        useShell,
       );
 
       await this.runBuild(
@@ -103,6 +106,7 @@ export class StartAction extends BuildAction {
     debugFlag: boolean | string | undefined,
     outDirName: string,
     binaryToRun: string,
+    useShell: boolean,
   ) {
     let childProcessRef: any;
     process.on(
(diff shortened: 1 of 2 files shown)


jeengbe:
Any update on this? 🙂  The EADDRINUSE issue makes local development rather inconvenient

quallrum:
@kamilmysliwiec is there smth we can do to get this PR merged?

micalevisk:
I'm not sure about the name 'no shell'. To me, it doesn't tell what will happen when using it. Do you know any other high level name? `--noShell` sound like we're exposing a detail implementation of `@nestjs/cli`. So if we want to change that implementation later, the flag will be out of sync somewhat

But I can't think of any other name tho

quallrum:
> I'm not sure about the name 'no shell'

I agree that the name isn't self-descriptive (though there is a description for that flag).

We can probably use smth like `--shell=false` since `spawn()`'s `shell` arg [supports different shells](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options) and we can either provide a custom shell (for whatever reason, just to be less specific then `noShell`) or disable it. 

> To be fair, I'm not sure if we should have `shell: true`

Me neither, i decided to introduce a flag just in favor of backward compatibility

ematthewsBW:
Thanks for this PR. Any update this? This will improve my DX.

pio-sko:
Phew! Finally PR has been nailed. I very glad to have this mode fixed on WSL.

PR #2726, opened 2024-09-07
fix(actions): ignore pkgmgr select when `--skip-install` is passsed

tbhaxor:
Since the scaffolder is emitting package.json only. Therefore, package manager prompt is not required here.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Even if we pass `-s` or `--skip-install` option, the CLI asks for the package manager. But now if these options are passed, the package manager selection prompt will be ignored.

```
[ ] Yes
[x] No
```

### DIFF
--- a/actions/new.action.ts
+++ b/actions/new.action.ts
@@ -94,6 +94,12 @@ const askForMissingInformation = async (inputs: Input[], options: Input[]) => {
     replaceInputMissingInformation(inputs, answers);
   }
 
+  // It should not ask for the package manager if --skip-install is option is set
+  const shouldSkipInstall = options.find((opt) => opt.name === 'skip-install')!;
+  if (shouldSkipInstall.value) {
+    return;
+  }
+
   const packageManagerInput = getPackageManagerInput(options);
   if (!packageManagerInput!.value) {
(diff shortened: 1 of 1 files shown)


micalevisk:
Hi

to be fair, the selected package manager is mentioned in the generated `README.md` file so this is expected. Not sure if we should drop this capability as we can use better in the future

tbhaxor:
Hi, I couldn't get what you meant. My PR is simple, it shouldn't prompt for selecting package manager if --skip-install is passed. Also, nest-cli.json file exists in the scaffolded project, not in global location, at least this is what I know.

micalevisk:
I meant that the selected package manager is important regardless of if you'll run `install` or not. Check this out: https://github.com/search?q=repo%3Anestjs%2Fschematics+%22%3C%25%3D+packageManager+%25%3E%22&type=code

tbhaxor:
It all makes sense now. Well, I am with you on nest-cli.json thing. There should be a global config, which is inherited (object merge) by the local one.

Closing the PR now as it doesn't make sense now.

PR #3128, opened 2025-08-15
fix(@nestjs/cli): disable shell execution on unix to prevent command injection

seoyeonjin:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

The CLI currently enables shell execution (`shell: true`) on all platforms when spawning child processes. This creates a command injection vulnerability on Unix-like systems (macOS/Linux) where shell metacharacters can be exploited to execute malicious commands.

Commands containing `$()`, `${}`, or other shell expansions can potentially execute arbitrary code on the system.

Issue Number: #3064 

## What is the new behavior?

Shell execution is now disabled on Unix-like systems (macOS/Linux) while remaining enabled on Windows for compatibility. This prevents command injection attacks on Unix systems while maintaining cross-platform functionality.

**Changes:**
- Unix systems: `shell: false` - prevents shell metacharacter exploitation
- Windows: `shell: true` - maintains compatibility
- Commands are safely passed as separate arguments to prevent injection

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #3193, opened 2025-11-19
fix: Node v24 DEP0190 warning by removing args for shell: true

kim-sung-jee:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Passing an args array together with `shell: true` triggers Node.js deprecation warning `DEP0190`

Issue Number: https://github.com/nestjs/nest/issues/15943

## What is the new behavior?

When shell is enabled, the CLI now passes a single command string to spawn instead of using the args array

This matches the pattern recommended by a Node.js member in
https://github.com/nodejs/help/issues/5063#issuecomment-3132899776

```
[ ] Yes
[x] No
```

## Other information

This change follows Node.js API guidance, rather than introducing a new escaping strategy.

Node.js deprecated the `args + shell: true` combination because the API shape gives a false impression that arguments are safely escaped, while they are in fact just concatenated into a single shell command string (see the DEP0190 change in [`normalizeSpawnArguments`](https://github.com/nodejs/node/pull/57199/files)).

This PR:

- Aligns the CLI with the recommended usage by passing a single command string when `shell: true`, and keeping `command + args` only when `shell` is `false` ([ the same pattern Node.js itself uses internally](https://github.com/nodejs/node/blob/934d90735ae36bc1bd362bf28f901d7b68c1176e/lib/child_process.js#L652) ).
- Avoids implying any “magic escaping” from the CLI side — we simply stop using the misleading `args + shell: true` combination.

Proper shell escaping / input hardening is intentionally **out of scope** for this PR and should be handled separately in a follow-up change, closer to where potentially untrusted input enters the system.

### DIFF
--- a/actions/start.action.ts
+++ b/actions/start.action.ts
@@ -1,5 +1,5 @@
 import { red } from 'ansis';
-import { spawn } from 'child_process';
+import { spawn, SpawnOptions } from 'child_process';
 import * as fs from 'fs';
 import { join } from 'path';
 import { Input } from '../commands';
@@ -218,9 +218,16 @@ export class StartAction extends BuildAction {
 
     processArgs.unshift('--enable-source-maps');
 
-    return spawn(binaryToRun, processArgs, {
+    const spawnOptions: SpawnOptions = {
       stdio: 'inherit',
       shell: options.shell,
-    });
+    };
+
+    if (options.shell) {
+      const command = [binaryToRun, ...processArgs].join(' ');
+      return spawn(command, spawnOptions);
+    }
+
+    return spawn(binaryToRun, processArgs, spawnOptions);
   }
 }
--- a/lib/runners/abstract.runner.ts
+++ b/lib/runners/abstract.runner.ts
@@ -22,11 +22,8 @@ export class AbstractRunner {
       shell: true,
     };
     return new Promise<null | string>((resolve, reject) => {
-      const child: ChildProcess = spawn(
-        `${this.binary}`,
-        [...this.args, ...args],
-        options,
-      );
+      const command = [this.binary, ...this.args, ...args].join(' ');
+      const child: ChildProcess = spawn(command, options);
       if (collect) {
         child.stdout!.on('data', (data) =>
           resolve(data.toString().replace(/\r\n|\n/, '')),


micalevisk (approved):
FYI this is not a `fix` but a `refactor`

kamilmysliwiec:
https://github.com/nestjs/nest-cli/pull/3192

This introduces another issue

kim-sung-jee:
@kamilmysliwiec @micalevisk  Thanks for the heads-up!

1. On Unix, `shell` used to split a single “command” string into proper argv tokens. After #3128, `shell` became `false` while the code still relied on the shell to do that splitting. For example, in `AbstractRunner`:

     ```ts
     // this.binary: 'node'
     // this.args: ['node_modules/@nestjs/schematics/bin/nest.js']
     // command: '@nestjs/schematics:application --name=practice --directory=... --package-manager="npm" ...'
    
     const args = [command];
    
     spawn(this.binary, [...this.args, ...args], {
       shell: false, // after #3128 on Unix
     });
    ```
    This results in:

    ```ts
    spawn('node', [
      'node_modules/@nestjs/schematics/bin/nest.js',
      '@nestjs/schematics:application --name=practice --directory=... --package-manager="npm" ...',
    ], { shell: false });
    ```

    Expected: 
    ```ts
    spawn('node', [
      'node_modules/@nestjs/schematics/bin/nest.js',
      '@nestjs/schematics:application',
      '--name=practice',
      '--directory=...',
      '--package-manager=npm',
      ...
    ], { shell: false });
   ```

    With `shell: true`, the shell split that long string into `@nestjs/schematics:application`, `--name=...`, `--directory=...`, etc. With `shell: false`, it’s passed as a single argv element, and schematics can’t parse it correctly.
    
    In addition, some arguments were already in a **shell-oriented format** (for example a quoted CLI path like `"path/to/schematics.js"` and a single `command` string with quoted options). With `shell: true`, the shell strips those quotes and does word splitting for us. Once `shell` is turned off but those formats stay the same, Node forwards the raw strings (including quotes) as argv elements, which is why `nest new` starts failing even though the underlying files and commands are correct.


2. **This PR only adapts to the Node.js API change for `shell: true`**  
   This PR doesn’t change when `shell` is enabled or disabled. It simply updates the `shell: true` call sites to match Node’s DEP0190 guidance: instead of `spawn(binary, args, { shell: true })`, the CLI now uses a single command string (`spawn(commandString, { shell: true })`), while the `shell: false` path stays as `spawn(binary, args, { shell: false })`.

Issue #3156, opened 2025-09-24
Replace abandoned tree-kill library (Issue with --watch on MacOS)


gcstarr:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

NestJS CLI uses `tree-kill` for its `--watch` functionality. The `tree-kill` library has been abandoned for 6 years and recently has started having issues running on MacOS v15+ (Sequoia and up), potentially more likely in mono-repos with large amounts of files (I'm unable to reproduce in a simple Hello World project). The issue is that when a change is detected, an EBADF error occurs when tree-kill tries to kill the running processes.

See [this issue in the tree-kill repo](https://github.com/pkrumins/node-tree-kill/issues/43), and [this conversation](https://discord.com/channels/520622812742811698/1364381581938262116) with yaacov and @micalevisk in NestJS's Discord server.

### Describe the solution you'd like

The main options are to use [one of the forks of tree-kill](https://www.npmjs.com/package/@magda/tree-kill) that has had this issue patched, of course risking that the forked library will also be abandoned, or to remove the external dependency entirely. Both approaches work, I've provided examples here and tested that both resolve my issues on MacOS Sequoia 15.5:

Use forked tree-kill example: https://github.com/nestjs/nest-cli/compare/master...gcstarr:nest-cli:use-tree-kill-fork

Remove external dependency example (was AI assisted as I was looking for a quick feasibility check, though I have tested it in both MacOS and Linux, but not Windows yet): https://github.com/nestjs/nest-cli/compare/master...gcstarr:nest-cli:remove-tree-kill

Thoughts?

### What is the motivation / use case for changing the behavior?

Removing abandoned dependency that is becoming more problematic.

gcstarr:
@kamilmysliwiec Now that I understand the code a bit better, I'm wondering if this just be as simple as: https://github.com/nestjs/nest-cli/commit/ba183c0e63be709e9f853e6d7fd7b0f457e05d48? The local tree-kill.ts you already have is using `execSync` which fixes the issue (The use of `spawn` in the original tree-kill library is what doesn't work well sometimes with newer MacOS builds).

kamilmysliwiec:
Yeah this should be enough indeed

eightHundreds:
I have also encountered this strange crash issue. I resolved it by downgrading from nodejs 20.19.5 to 20.18.0.

gcstarr:
[...]

PR #3235, opened 2026-01-08
fix(utils): tree-kill use spawn sync to fix node 24 crash

LucM:
## PR Checklist

* [x] The commit message follows our guidelines: [https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features)
* [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

On **Node.js v24+ (specifically on macOS)**, running the CLI in watch mode (e.g., `nest start --watch`) crashes when a file change is detected.

The error occurs in `getAllPid` because `execSync` attempts to spawn a shell (`/bin/sh`) to run `ps`. Newer versions of Node.js have stricter handling of file descriptor inheritance during process reloading, causing the shell invocation to fail with `EBADF`.

**Error Log:**

```
Error: spawnSync /bin/sh EBADF
    at Object.spawnSync (node:internal/child_process:1120:20)
    at execSync (node:child_process:991:15)
    at getAllPid (.../lib/utils/tree-kill.js:...)

```

Issue Number: #3156 

## What is the new behavior?

The `getAllPid` function now uses `spawnSync` directly instead of `execSync`.

1. **Bypasses Shell:** By invoking the `ps` binary directly (instead of via `/bin/sh`), we avoid the shell-related file descriptor issues on Node 24.
2. **Explicit Stdio:** We explicitly set `stdio: 'pipe'` to prevent invalid file descriptor inheritance.
3. **Safety:** Added a `try/catch` block. If `ps` fails for any system-level reason, the CLI now defaults to an empty list (no processes found) rather than crashing the entire watch process.

```
[ ] Yes
[x] No
```

## Other information

* Tested on macOS with Node.js v24.4.0.
* Confirmed that `nest start --watch` correctly restarts the application after file changes without crashing.

### DIFF
--- a/lib/utils/tree-kill.ts
+++ b/lib/utils/tree-kill.ts
@@ -1,4 +1,4 @@
-import { execSync } from 'child_process';
+import { execSync, spawnSync } from 'child_process';
 
 export function treeKillSync(pid: number, signal?: string | number): void {
   if (process.platform === 'win32') {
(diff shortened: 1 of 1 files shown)


shnooshnoo:
any ETA on when this may be available? Consistently having this issue, any change causes a crash.

PR #3048, opened 2025-05-29
feat: pass exclude from tsconfig to swc compiler

fjodor-rybakov:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: https://github.com/nestjs/nest-cli/issues/2976

## What is the new behavior?

Pass `exclude` from `tsconfig` to `--ignore` flag
https://swc.rs/docs/usage/cli#--ignore

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

## Other information

Issue #2976, opened 2025-04-04
swc+watch cannot exclude files


fpnuseis:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

if use swc + watch option, can not exclude files(ex. *spec.ts)

### Describe the solution you'd like

add watch exclude option for swc

### What is the motivation / use case for changing the behavior?

When modifying *spec.ts files, these files are generally excluded because we don't want a rebuild.
By adding an exclude option for swc, a rebuild can be avoided when certain files are changed.

kamilmysliwiec:
Would you like to create a PR for this issue?

Antraxmin:
I'd like to work on this issue. I'll create a PR that implements the watch exclude option for swc.

fpnuseis:
> 저는 이 문제에 대해 작업하고 싶습니다. swc에 대한 watch exclude 옵션을 구현하는 PR을 만들겠습니다.

Thanks!

fpnuseis:
> I'd like to work on this issue. I'll create a PR that implements the watch exclude option for swc.

Thanks!

fjodor-rybakov:
I'm impemented this feature. Can you review it?

PR #3276, opened 2026-02-19
feat: respect tsconfig excludes when using swc

enduity:
## PR Checklist

    - I am unsure where/how to add these. I would be happy to do so after receiving additional guidance.

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When switching a fresh [NestJS starter](https://github.com/nestjs/typescript-starter/tree/c4d9330f5513eda0fb5df594f6b34a11fde1a934) over to use SWC (using the official [guide](https://docs.nestjs.com/recipes/swc#swc)), the `exclude` field in `tsconfig.build.json` is no longer considered for build output. This means that a file like `src/app.controller.spec.ts` (included in the starter) ends up in the build as `dist/app.controller.spec.js`, even though `tsconfig.build.json` contains an exclusion glob pattern of `"**/*spec.ts"`.

It is expected that switching the NestJS builder from `tsc` to `swc` does not cause test files to be included in the `nest build` output.

The same concern applies in watch mode for newly added files, because Nest CLI manually watches source additions for SWC and triggers compilation for added files.

Issue Number: #2976

## What is the new behavior?

The SWC builder now uses `exclude` values from the parsed TypeScript config to populate SWC CLI `ignore` defaults.

This includes:

- exposing normalized `exclude` values from `TsConfigProvider`
- supporting inherited `exclude` values from `tsconfig` `extends`
- passing excludes to SWC CLI `ignore`
- respecting those ignore patterns when SWC watch mode manually compiles newly added source files
- preserving user-provided SWC builder option overrides

```text
[ ] Yes
[x] No
```

Unfortunately, it is _possible_ that someone has come to rely on the unexpected behaviour of `tsconfig.build.json` excludes not working for SWC compilation. However, I do think it's extremely unlikely. I welcome opposing thoughts.

## Caveats

This PR forwards `tsconfig` `exclude` values to SWC CLI `ignore` defaults so SWC's directory traversal does not emit files that are commonly excluded from `tsconfig.build.json`, such as `*.spec.ts` files.

It does not try to provide full TypeScript program parity. TypeScript's `exclude` only affects root file discovery, and excluded files can still be included if imported by included files. The SWC builder already uses a source-directory traversal model, so this change stays within that model instead of trying to reconstruct TypeScript's full file graph.

Explicit SWC builder options continue to override CLI defaults. That means `compilerOptions.builder.options.ignore`, when provided, remains authoritative rather than being force-merged with excludes from `tsconfig`.

## Tests

Added/updated tests for:

- parsing and normalizing `exclude` from TypeScript config
- inherited `exclude` values through `extends`
- SWC defaults mapping `exclude` to `cliOptions.ignore`
- SWC watch-mode handling for newly added ignored files
- absolute path / glob escaping behavior for ignore matching

## Other information

Also, there was already a similar PR, but it used potentially more fragile mechanisms to retrieve the `exclude` value from a `tsconfig.json`:
- https://github.com/nestjs/nest-cli/pull/3048

### DIFF
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -1,7 +1,6 @@
 import { createRequire } from 'module';
 import { red } from 'ansis';
 import { join } from 'path';
-import type * as ts from 'typescript';
 import { BuildCommandContext } from '../commands/index.js';
 import { AssetsManager } from '../lib/compiler/assets-manager.js';
 import { deleteOutDirIfEnabled } from '../lib/compiler/helpers/delete-out-dir.js';
@@ -11,7 +10,10 @@ import { getRspackConfigPath } from '../lib/compiler/helpers/get-rspack-config-p
 import { getTscConfigPath } from '../lib/compiler/helpers/get-tsc-config.path.js';
 import { getValueOrDefault } from '../lib/compiler/helpers/get-value-or-default.js';
 import { getWebpackConfigPath } from '../lib/compiler/helpers/get-webpack-config-path.js';
-import { TsConfigProvider } from '../lib/compiler/helpers/tsconfig-provider.js';
+import {
+  TsConfigProvider,
+  TsConfigProviderOutput,
+} from '../lib/compiler/helpers/tsconfig-provider.js';
 import { PluginsLoader } from '../lib/compiler/plugins/plugins-loader.js';
 import { TypeScriptBinaryLoader } from '../lib/compiler/typescript-loader.js';
 import {
@@ -214,7 +216,7 @@ export class BuildAction extends AbstractAction {
     pathToTsconfig: string,
     watchMode: boolean,
     options: Record<string, any>,
-    tsOptions: ts.CompilerOptions,
+    tsOptions: TsConfigProviderOutput['options'],
     emitDeclarations: boolean,
     onSuccess: (() => void) | undefined,
   ) {
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -85,8 +85,11 @@ export class BuildAction extends AbstractAction {
 
     const buildApp = async (appName: string | undefined) => {
       const pathToTsconfig = getTscConfigPath(configuration, options, appName);
-      const { options: tsOptions, fileNames: tsFileNames } =
-        this.tsConfigProvider.getByConfigFilename(pathToTsconfig);
+      const {
+        exclude,
+        options: tsOptions,
+        fileNames: tsFileNames,
+      } = this.tsConfigProvider.getByConfigFilename(pathToTsconfig);
       const outDir = tsOptions.outDir || defaultOutDir;
       const tsRootDir = getEffectiveRootDir(tsOptions.rootDir, tsFileNames);
 
@@ -181,6 +184,7 @@ export class BuildAction extends AbstractAction {
             watchMode,
             options,
             tsOptions,
+            exclude,
             emitDeclarations,
             onSuccess,
           );
@@ -217,6 +221,7 @@ export class BuildAction extends AbstractAction {
     watchMode: boolean,
     options: Record<string, any>,
     tsOptions: TsConfigProviderOutput['options'],
+    tsConfigExclude: string[],
     emitDeclarations: boolean,
     onSuccess: (() => void) | undefined,
   ) {
@@ -239,6 +244,7 @@ export class BuildAction extends AbstractAction {
         ),
         emitDeclarations,
         tsOptions,
+        tsConfigExclude,
         assetsManager: this.assetsManager,
         silent: isSilent,
       },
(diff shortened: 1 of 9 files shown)


kamilmysliwiec:
Could you rebase to v12.0.0?

kamilmysliwiec:
Just checking if you're still interested in working on this PR

enduity:
> Just checking if you're still interested in working on this PR

Yeah, sorry, I did not have a lot of time available since your request for me to rebase this. Working on it now.

actions/build.action.ts
```diff
@@ -25,6 +24,7 @@ import { ERROR_PREFIX, INFO_PREFIX } from '../lib/ui';
 import { isModuleAvailable } from '../lib/utils/is-module-available';
 import { AbstractAction } from './abstract.action';
 import webpack = require('webpack');
+import { TsConfigProviderOutput } from '../lib/compiler/helpers/tsconfig-provider';
```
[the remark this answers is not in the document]
enduity: thanks
31069cbd

test/lib/compiler/helpers/tsconfig-provider.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+      (fs.readFileSync as jest.Mock).mockReturnValue(tsconfigContent);
+      typescriptLoaderMock.load.mockRestore();
+
+      const result = provider.getByConfigFilename(configFilename);
+
+      expect(result.options.outDir).toContain('dist');
+      expect(result.options.exclude).toEqual(['node_modules', 'dist']);
+      expect(result.fileNames).toBeDefined();
+
+      expect(fs.readFileSync).toHaveBeenCalledWith(configPath);
```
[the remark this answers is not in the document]
enduity: thanks
13bdb7d6

lib/compiler/helpers/tsconfig-provider.ts
```diff
@@ -4,10 +4,29 @@ import * as ts from 'typescript';
...
+  'fileNames' | 'projectReferences'
+> & {
+  options: ts.ParsedCommandLine['options'] & { exclude: string[] };
+};
+
 export class TsConfigProvider {
   constructor(private readonly typescriptLoader: TypeScriptBinaryLoader) {}
 
-  public getByConfigFilename(configFilename: string) {
+  private parseExclude(exclude: unknown): string[] {
```
kamilmysliwiec: could we move to the end of the class? or at least below public methods

enduity: Sorry, I forgot to change this PR to be a draft, this was still a work-in-progress.

Fixup:
- 5f3bbb83

Thanks for the feedback!

enduity:
Small status update: the original implementation was fairly small, but during the rebase I noticed the SWC watch-mode path did not go through the same exclude handling for newly added files. I added coverage for that path as well, so the PR is larger than before.

I have converted this back to draft while I finish validation and cleanup. It should be coherent enough for early direction-level feedback, but I will mark it ready once the branch is fully polished and checks are in a good state.

kamilmysliwiec:
Excellent, thank you

enduity:
I polished things up and added a "Caveats" section to the PR description. This still is not perfect TypeScript parity, but I think it matches the common expectation for `tsconfig.build.json`: files like specs should not end up in `dist` after `nest build`.

The main case where behavior could differ from `tsc` is if a source file imports another file under `src/` that is excluded from `tsconfig.build.json` for some reason. That seems pretty unusual, and users who need that can still override the SWC ignore configuration explicitly.

package.json
```diff
@@ -50,6 +50,7 @@
     "cli-table3": "0.6.5",
     "commander": "14.0.3",
     "glob": "13.0.6",
+    "minimatch": "10.2.5",
```
kamilmysliwiec, line 53: now we have 2 glob-like libraries as dependencies

enduity: It is already in the dependency tree as a `glob` subdependency

https://github.com/nestjs/nest-cli/blob/10a59f9c160dbdb79890b847d141fb25b4533161/package-lock.json#L6932-L6940

This just exposes `minimatch` for direct use.

The problem I was solving is that `glob` is only meant for globbing an actual filesystem, making it impractical to use for [checking](https://github.com/enduity/nest-cli_fork/blob/aea20a477c6c9b308a1c05b69261a6bbefd253f7/lib/compiler/swc/swc-compiler.ts#L371) if a string _path_ matches a specific glob pattern.

It is also what swc-cli uses internally:
- https://github.com/swc-project/pkgs/blob/d20fdcb6e1268e27295ccaa28428a6f25e5e9c78/packages/cli/src/swc/sources.ts#L55

kamilmysliwiec: fair enough!

Issue #2847, opened 2024-12-16
Big package size due to un-deduplicated copy of typescript


golopot:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Currently the effective package size of `@nest/cli` is 34MB, mainly because the package requires a pinned version of typescript, which make npm/yarn unable to deduplicate typescript package. Can we loosen the version requirement for typescript?

```
% du -sh node_modules/@nestjs/cli | sort -h 
 34M    node_modules/@nestjs/cli
% du -sh node_modules/@nestjs/cli/node_modules/* | sort -h
164K    node_modules/@nestjs/cli/node_modules/cosmiconfig
740K    node_modules/@nestjs/cli/node_modules/fork-ts-checker-webpack-plugin
1.1M    node_modules/@nestjs/cli/node_modules/glob
 31M    node_modules/@nestjs/cli/node_modules/typescript
```

https://github.com/nestjs/nest-cli/blob/95c7f7e91521125d39997a4401794dd4c0a0e5a1/package.json#L57-L57

### Describe the solution you'd like

 Change typescript version to something like "^5.7.2". 

### What is the motivation / use case for changing the behavior?

Solving this can reduce network cost and build time in local development and CI.  This package `@nest/cli` has about 2 million downloads per week, saving 30MB in size would be 60TB / per week of transfer cost, which is about $3000 of CDN hosting cost per week.

micalevisk:
I think that we should use `~` instead of `^` because, if I'm not wrong, sometimes `typescript` bumps the minor slice that might introduce breaking changes for compiler's API (which Nest's CLI uses under the hood)

PR #3339, opened 2026-04-09
fix(deps): use tilde range for typescript version

maruthang:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `typescript` dependency in `package.json` is pinned to an exact version (`5.9.3`). This prevents npm/yarn from deduplicating it with the project's own typescript installation, resulting in a duplicate copy in `node_modules` and approximately 31 MB of unnecessary disk usage.

Issue Number: #2847

## What is the new behavior?

The `typescript` dependency now uses a tilde range (`~5.9.3`), which allows the package manager to deduplicate typescript at the patch level. This avoids a redundant copy when the project already has a compatible typescript version installed. A tilde range (rather than caret) is used because TypeScript can introduce breaking compiler API changes in minor version bumps.

```
[ ] Yes
[x] No
```

## Other information

- Added regression test in `test/package-json.spec.ts` to ensure typescript uses a tilde range and is not pinned to an exact version.


maruthang:
Hi @kamilmysliwiec, CI is now green — the failing check was a vitest compatibility issue in `test/package-json.spec.ts` (missing explicit imports of `describe`/`it`/`expect`/`beforeAll` from vitest, which the v12.0.0 migration requires). Rebased onto `v12.0.0` and all checks pass. Ready for another look whenever you have a moment — thanks!

Issue #2626, opened 2024-07-03
Add option to restart Application on Asset change


MarjanDB:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

The problem I'm facing is similar to these two already existing closed issues:
- https://github.com/nestjs/nest-cli/issues/1055
- https://github.com/nestjs/nest-cli/issues/1136

I have definition / configuration files that are read by the application during its bootstrap phase which configures certain features of the application. Using the `watchAssets` option [as described here](https://docs.nestjs.com/cli/monorepo#assets) does update the files in the distribution folder when they're changed, but since they're only picked up during the application bootstrap phase, this effectively does nothing (as the application requires a restart for the changes to take effect). I have to resort to manually restarting the server, or if the application is running inside a mounted docker container, changing a random bit of code so that `nest start --watch` restarts the application.

### Describe the solution you'd like

Changing the behavior of `watchAssets` to do this would be nice, but I understand that this sort of change is not something others developers may want.

Since this is the case, introducing a new option to handle the restart could be a viable solution to the problem that does not break existing workflows.

An option along the lines of `restartApplicationOnChange` could be added wherever `watchAssets` is currently possible to be specified. When this option is set to `true`, the application will restart if the asset is changed.

This would also give the developer the ability to more finely control when the application requires a restart should they need it.

Something that would fall out with this change would be the inconsistency of "watch" options :
-  `--watch` **will restart** on code changes,
- `watchAssets` **will not restart** on asset changes,
- but specifying `restartApplicationOnChange` alongside `watchAssets` **will restart** on asset changes

Perhaps `restartApplicationOnChange` could be considered as being false by default, and later changed to `true` by default with a major release. This would resolve the inconsistency of "watch" options while still providing a method for developers to go back to using their old workflows (no restarts on asset changes).

### What is the motivation / use case for changing the behavior?

Already explained in the first section, but in short:
I have configuration files that are read during bootstrap and only during bootstrap. A restart of the application is required in order to actually apply the changes that were made.

micalevisk:
doesn't typing `rs` (with `compilerOptions.manualRestart:true`) resolves this?

MarjanDB:
@micalevisk 
> doesn't typing `rs` (with `compilerOptions.manualRestart:true`) resolves this?
If running on the local machine, yes (with the caveat that you have to manually do it).
[...]

PR #3340, opened 2026-04-09
feat(assets): restart app on asset change in watch mode (#2626)

maruthang:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When running `nest start --watch --watchAssets`, changes to TypeScript files trigger an application restart, but changes to watched asset files (e.g. `.proto`, `.graphql`) only copy the updated file without restarting the application. This means developers must manually restart to pick up asset changes.

Issue Number: #2626

## What is the new behavior?

Asset file changes now trigger an application restart when using `--watch --watchAssets`. This is achieved by adding an optional `onSuccess` callback parameter to `AssetsManager.copyAssets()`. The `BuildAction` passes its existing `onSuccess` handler (the one that restarts the app) to `copyAssets()`, so asset changes invoke the same restart logic as TypeScript changes. The callback is only invoked after chokidar emits the `ready` event, preventing unnecessary restarts during the initial file scan.

```
[ ] Yes
[x] No
```

## Other information

**Files changed:**
- `lib/compiler/assets-manager.ts` — added optional `onSuccess` callback parameter, invoked on post-ready add/change/unlink events
- `actions/build.action.ts` — passes `onSuccess` callback through to `copyAssets()`
- `test/lib/compiler/assets-manager.spec.ts` — added 4 regression tests covering callback invocation, no-callback during initial scan, and no-callback when not provided

### DIFF
--- a/lib/compiler/assets-manager.ts
+++ b/lib/compiler/assets-manager.ts
@@ -11,6 +11,8 @@ import {
 import { copyPathResolve } from './helpers/copy-path-resolve.js';
 import { getValueOrDefault } from './helpers/get-value-or-default.js';
 
+const ASSET_CHANGE_RESTART_DEBOUNCE_MS = 150;
+
 export class AssetsManager {
   private watchAssetsKeyValue: { [key: string]: boolean } = {};
   private watchers: chokidar.FSWatcher[] = [];
@@ -32,6 +34,7 @@ export class AssetsManager {
     appName: string | undefined,
     outDir: string,
     watchAssetsMode: boolean,
+    onSuccess?: () => void,
   ) {
     const assets =
       getValueOrDefault<Asset[]>(
@@ -74,6 +77,22 @@ export class AssetsManager {
           appName,
         ) || watchAssetsMode;
 
+      // Debounce onSuccess so that a burst of asset changes (e.g. a git
+      // checkout touching many files at once) only triggers a single restart.
+      let debouncedOnSuccess: (() => void) | undefined;
+      if (onSuccess) {
+        let pending: NodeJS.Timeout | undefined;
+        debouncedOnSuccess = () => {
+          if (pending) {
+            clearTimeout(pending);
+          }
+          pending = setTimeout(() => {
+            pending = undefined;
+            onSuccess();
+          }, ASSET_CHANGE_RESTART_DEBOUNCE_MS);
+        };
+      }
+
       for (const item of filesToCopy) {
         const option: ActionOnFile = {
           action: 'change',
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
Could you rebase to v12.0.0?

JSteunou:
Does this mean it wont be release in current v11?

maruthang:
Hi @JSteunou, good question — since @kamilmysliwiec asked to rebase this onto `v12.0.0`, it looks like this feature is targeting the v12 release. That said, backporting to v11 would be up to the maintainers' discretion. Might be worth opening a separate discussion/issue if v11 support is important for your use case!

lib/compiler/assets-manager.ts
```diff
@@ -84,15 +85,35 @@ export class AssetsManager {
...
               ignore: item.exclude,
               dot: true,
             }))
-            .on('add', (path: string) => this.actionOnFile({ ...option, path, action: 'change' }))
-            .on('change', (path: string) => this.actionOnFile({ ...option, path, action: 'change' }))
-            .on('unlink', (path: string) => this.actionOnFile({ ...option, path, action: 'unlink' }));
+            .on('add', (path: string) => {
+              this.actionOnFile({ ...option, path, action: 'change' });
+              if (ready && onSuccess) {
+                onSuccess();
```
kamilmysliwiec: shouldn't we debounce this call a little, to avoid, multiple restarts in case more than 1 file is updated at the same time?

maruthang: Good call — added a 150ms trailing debouncer around the `onSuccess` call so bursts of file events (git checkouts, bulk saves, etc.) collapse into a single restart. Also rebased onto the latest `v12.0.0` to clear the merge conflict, and added a regression test `should debounce rapid onSuccess calls into a single invocation` that fires four events in quick succession and asserts only one restart callback fires.

All 11 assets-manager unit tests pass locally. Ready for another look!

kamilmysliwiec:
lgtm

JSteunou:
> Hi @JSteunou, good question — since @kamilmysliwiec asked to rebase this onto `v12.0.0`, it looks like this feature is targeting the v12 release. That said, backporting to v11 would be up to the maintainers' discretion. Might be worth opening a separate discussion/issue if v11 support is important for your use case!

I copied most of this PR through a pnpm patch, it does the job until v12 release.

@kamilmysliwiec no worries, I can understand if you prefer to move fast on v12 and not spend time on v11 except for important fixes.

PR #3342, opened 2026-04-09
feat(swc): add emit declarations flag for type definition generation

### LINKED ISSUE #2610: Allow to generate type declaration files with `tsc` when the builder is `swc` or to skip building in `nest start`

It seems when using `swc`, we cannot generate the `*.d.ts` files [[1](https://github.com/swc-project/swc/issues/657),[2](https://github.com/swc-project/swc/issues/571)] needed by, e.g., MikroOrm's `TsMorphMetadataProvider` [[3](https://mikro-orm.io/docs/metadata-providers#tsmorphmetadataprovider),[4](https://github.com/mikro-orm/mikro-orm/blob/master/packages/reflection/src/TsMorphMetadataProvider.ts#L30-L38)]. One of the workarounds to call `tsc -p tsconfig.declaration.json`, with
```javascript
// tsconfig.declaration.json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "emitDeclarationOnly": true,
    "declaration": true,
    "declarationDir": "./dist"
  }
}
```
after the building step. 

However, when using `nest start`, this will require manually splitting the `start` command into:
1. Setting `"deleteOutDir": false` in the `compilerOptions` of `nest-cli.json`
2. `rm -rf ./dist`
3. `nest build`
4.  `tsc -p tsconfig.declaration.json`
5. `nest start`

which seems cumbersome.

It would be convenient if we could do something like `nest --generate-declarations start` or `nest build && tsc -p tsconfig.declaration.json && nest --skip-builder start` (or `nest --builder none start`). We may add some checks before the [`this.runBuild` line in `start.action.ts`](https://github.com/nestjs/nest-cli/blob/master/actions/start.action.ts#L83) to achieve this.

_No response_

See the first section.

maruthang:
## Summary

- Adds `--emit-declarations` CLI flag to `nest build` and `nest start` commands
- When using SWC builder, runs `tsc --emitDeclarationOnly` after SWC compilation to generate `.d.ts` files
- Supports both CLI flag and `compilerOptions.emitDeclarations` in `nest-cli.json`
- Shows a warning when used with non-SWC builders (tsc already emits declarations)

Closes #2610

## Test plan

- [x] Added 5 new tests for emit declarations feature in swc-compiler.spec.ts
- [x] All 156 existing tests continue to pass
- [ ] Manual test: `nest build --builder swc --emit-declarations` generates .d.ts files
- [ ] Manual test: `nest start --builder swc --emit-declarations` generates .d.ts during start

### DIFF
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -119,6 +119,20 @@ export class BuildAction extends AbstractAction {
         );
       }
 
+      const emitDeclarations = getValueOrDefault<boolean>(
+        configuration,
+        'compilerOptions.emitDeclarations',
+        appName,
+        'emitDeclarations',
+        commandOptions,
+      );
+      if (emitDeclarations && builder.type !== 'swc') {
+        console.warn(
+          INFO_PREFIX +
+            ` "emitDeclarations" will not have any effect when "builder" is not "swc".`,
+        );
+      }
+
       switch (builder.type) {
         case 'tsc':
           await this.runTsc(
@@ -160,6 +174,7 @@ export class BuildAction extends AbstractAction {
             watchMode,
             options,
             tsOptions,
+            emitDeclarations,
             onSuccess,
           );
           break;
@@ -174,6 +189,7 @@ export class BuildAction extends AbstractAction {
     watchMode: boolean,
     options: Record<string, any>,
     tsOptions: ts.CompilerOptions,
+    emitDeclarations: boolean,
     onSuccess: (() => void) | undefined,
   ) {
     const { SwcCompiler } = await import('../lib/compiler/swc/swc-compiler.js');
@@ -193,6 +209,7 @@ export class BuildAction extends AbstractAction {
           'typeCheck',
           options,
         ),
+        emitDeclarations,
         tsOptions,
         assetsManager: this.assetsManager,
         silent: isSilent,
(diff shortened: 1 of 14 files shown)


maruthang:
Hi @kamilmysliwiec, CI fix pushed — the failing unit test (`info.action.spec.ts`) was still using `jest.spyOn` which isn't available under vitest in v12.0.0. Swapped to `vi.spyOn`. All tests pass locally. Thanks for the review!

kamilmysliwiec:
could you add an e2e test?

maruthang:
Added e2e coverage for `--emit-declarations` in `test/e2e/build.command.e2e-spec.ts` within the existing `Build Command (e2e) > with SWC builder` block:

- **Positive**: `nest build --builder swc --emit-declarations` — asserts `.d.ts` files exist for `app.module`, `app.controller`, and `app.service` alongside the compiled JS.
- **Negative**: `nest build --builder swc` (no flag) — asserts `.d.ts` files are **not** emitted by default.

Reusing the existing scaffolded app and the SWC deps that are already installed in the parent `beforeAll`, so no extra setup cost. Ready for another look!

maruthang:
Follow-up on the e2e test - the flag was recognized after `removeLocalCli`, but the `.d.ts` files weren't showing up because the feature delegates to `tsc --emitDeclarationOnly`, which silently produces no output when `declaration: false` (the default in the scaffolded `tsconfig.build.json`).

Updated the test to flip `declaration: true` before invoking the build, and restore the original tsconfig in a `finally` block so subsequent tests in the same spec aren't affected. Fresh CI triggered.

kamilmysliwiec:
Could you create a PR to the docs as well?

PR #3343, opened 2026-04-09
feat(assets): add option to include library assets in app builds

### LINKED ISSUE #1845: Assets feature is silently ignored in libraries - should either fail or work

A minimum viable repro of this issue is available with further description here: https://github.com/zackdotcomputer/nest-libs-assets-repro

The issue here is as follows:
1. If I create a library, I can specify `compilerOptions.assets` for that library and this works as expected when building that library by itself.
2. However, if I use that library inside of an application, the library's assets options are silently ignored. The code for the library is incorporated into the application (i.e. even if you build the library separately to get assets working, you cannot then use that copy of the library) and the assets are not.

At minimum, whatever is decided here should be noted on the pages documenting [assets](https://docs.nestjs.com/cli/monorepo#assets) and [libraries](https://docs.nestjs.com/cli/libraries) since currently there is no way of knowing about this limitation before developing for it.

As for an in-code resolution, any of the following options would, in my opinion, be satisfactory. My preference, however, is for number 3.

1. **Fail the compilation if the user has specified assets inside of a library used by the application.**
  - Con: this would mean that builds might fail where they did not previously.
  - Pros: this means fewer builds will appear to succeed but actually contain a behavior that is not the developer's intention (i.e. I suspect most users who specify assets inside of a library expect those assets to be incorporated in the built product, so the CLI should flag that that is not the case).

2. **Include all assets in libraries inside of built applications.** This would require having `assets-manager.ts` include all `compilerOptions.assets`-specified files into the output folders for application(s) built by `nest build`.
  - Con: this might be overzealous - currently the assets manager cannot ascertain which libraries an application depends on so it may include assets that a given application does not actually require.
  - Pro: this would enable users to define assets in libraries and use them in applications.

3. **Allow an application's assets list to reference other libraries.** This would entail a change to `assets-manager.ts` so that a reference could be made to libraries whose assets should be included. An example syntax might be `"assets": { "libraries": ["images", "schemas"] }`.
  - Con: This requires an extension to the `nest-cli.json` schema and thus might be the largest lift in work.
  - Pro: This is the best solution IMO because it does not change the default behavior but enables the desired behavior of being able to silo assets into reusable libraries.

If my proposed solution (3) above were approved, I'd suggest that we separate the documentation of Assets from the Monorepo page where it currently is. We would then want to make the following two changes:
1. Expand the example section at top so that it's clearer how and where the assets object/array is being defined.
2. Add a new section title "Assets in Libraries" that explains how to specify dependencies on assets in libraries.

Solution 3 requires no direct migrations strategy because it does not change existing behavior. It only adds support for new functionality.

I had been attempting to use the existing assets feature in `nest build` to avoid having to make repetitive changes to the build pipelines for multiple applications produced by a single Nest monorepo. There are several assets in a shared library which must be included in the final output for each application. In this case they are the schema and client binaries for Prisma, which Nest's sample repo recommends adding [item-by-item COPY statements](https://github.com/notiz-dev/nestjs-prisma-starter/blob/main/Dockerfile.alpine) for to one's Dockerfile, but the same principal applies to any such set of assets in a library which could be useful in multiple outputs.

maruthang:
## Summary

- Adds `includeLibraryAssets` option to `compilerOptions` in `nest-cli.json`
- Accepts an array of library project names whose assets should be included when building an application
- Resolves library asset paths relative to each library's `sourceRoot`
- Non-existent library names and libraries without assets are gracefully skipped

Example configuration:
```json
{
  "compilerOptions": {
    "includeLibraryAssets": ["my-library", "shared-schemas"]
  }
}
```

Closes #1845

## Test plan

- [x] Added 6 new tests covering: library asset inclusion, multiple libraries, missing libraries, libraries without assets
- [x] All 157 tests pass
- [ ] Manual test: monorepo with library assets included in application build output

### DIFF
--- a/lib/compiler/assets-manager.ts
+++ b/lib/compiler/assets-manager.ts
@@ -43,7 +43,20 @@ export class AssetsManager {
         appName,
       ) || [];
 
-    if (assets.length <= 0) {
+    const includeLibraryAssets =
+      getValueOrDefault<string[]>(
+        configuration,
+        'compilerOptions.includeLibraryAssets',
+        appName,
+      ) || [];
+
+    const libraryAssets = this.collectLibraryAssets(
+      configuration,
+      includeLibraryAssets,
+      outDir,
+    );
+
+    if (assets.length <= 0 && libraryAssets.length <= 0) {
       return;
     }
 
@@ -70,6 +83,8 @@ export class AssetsManager {
         };
       });
 
+      const allFilesToCopy = [...filesToCopy, ...libraryAssets];
+
       const isWatchEnabled =
         getValueOrDefault<boolean>(
           configuration,
@@ -93,12 +108,13 @@ export class AssetsManager {
         };
       }
 
-      for (const item of filesToCopy) {
+      for (const item of allFilesToCopy) {
+        const itemSourceRoot = (item as any)._sourceRoot || sourceRoot;
         const option: ActionOnFile = {
           action: 'change',
           item,
           path: '',
-          sourceRoot,
+          sourceRoot: itemSourceRoot,
           watchAssetsMode: isWatchEnabled,
         };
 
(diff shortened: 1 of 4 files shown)


maruthang:
Sure, rebasing to v12.0.0 and adding e2e tests now!

maruthang:
Hi @kamilmysliwiec, rebased onto latest `v12.0.0` to pick up recent upstream changes, and re-verified locally — all 456 unit tests pass and the e2e tests for library asset inclusion (`test/e2e/library-assets.e2e-spec.ts`) pass as well. Fresh CI run triggered by the push. Ready for another look — thanks!

maruthang:
Fixed the e2e CI failure — the test was scaffolding a monorepo with `webpack: true` in `nest-cli.json` but not installing webpack peer deps, causing the build to fail. The `includeLibraryAssets` feature doesn't need webpack, so I disabled it in the test to use the default tsc compiler. Fresh CI run triggered.

kamilmysliwiec:
e2e tests still fail

maruthang:
Thanks for the heads-up @kamilmysliwiec — I fetched the CircleCI log and found the root cause: the scaffold helper (`scaffoldMonorepoWithDeps`) uses a naive `str[0].toUpperCase() + str.slice(1)` transform to name classes, which produces invalid TypeScript identifiers like `My-libModule` when the sub-app name has a hyphen. My test was passing `'my-lib'`.

Just renamed the sub-app to `mylib` in the test — the generated `MylibModule`/`MylibService`/`MylibController` compile cleanly. Fresh CI run triggered. (Happy to also fix the helper to properly PascalCase hyphenated names in a follow-up PR if you'd like.)

maruthang:
Another CI fix pushed — two issues were stacked:

1. **ESM/CJS mismatch**: `scaffoldMonorepoWithDeps` scaffolds an ESM project (`nest new` defaults to `type: module` in v12), but the helper's manually-written sub-app files use CJS-style imports without `.js` extensions — so with `moduleResolution: NodeNext` the build failed with `TS2307: Cannot find module './mylib.module'`. Added `convertToCjs(appPath)` like the existing webpack monorepo test does.

2. **Published CLI doesn't know about `includeLibraryAssets`**: `scaffoldAppWithDeps` does an `npm install` which pulls the published `@nestjs/cli`, and `bin/nest.js` delegates to the local install via `localBinExists()` — so the dev code for this feature was never being exercised. Added `removeLocalCli(appPath)`.

Fresh CI run triggered.

maruthang:
One more fix — the assertions expected `dist/apps/<name>/src/main.js` but with `include: ['src/**/*']` tsc infers `rootDir` as `src/` so the output lands directly at `dist/apps/<name>/main.js` (same path the existing webpack monorepo test in `build.command.e2e-spec.ts` asserts). Fresh CI triggered.

kamilmysliwiec:
Could you resolve conflicts?

maruthang:
Conflicts resolved by rebasing onto latest `v12.0.0`. Combined the new `includeLibraryAssets` option with the upstream debounce/declarations changes; updated existing `closeWatchers` and `onSuccess` tests to mock the additional `getValueOrDefault` call. Local unit tests pass (the 4 pre-existing `tsconfig-paths.hook.spec.ts` failures also reproduce on clean `v12.0.0` and are unrelated).

kamilmysliwiec:
Could you create a PR to the docs as well?

maruthang:
Companion docs PR opened: https://github.com/nestjs/docs.nestjs.com/pull/3406

PR #3345, opened 2026-04-09
fix(tsconfig-provider): add null check for parsed config result

yogeshwaran-c:

Bug fix

`TsConfigProvider.getByConfigFilename()` uses a non-null assertion (`parsedCmd!`) on the return value of `getParsedCommandLineOfConfigFile()`. When this function returns `undefined` (e.g., due to invalid JSON or unsupported compiler options in the tsconfig file), the destructuring crashes with a confusing error:

```
Cannot destructure property 'options' of undefined
```

## What is the new behavior?

Adds an explicit null check before destructuring `parsedCmd`. When parsing returns `undefined`, a descriptive error is thrown:

```
Could not parse TypeScript configuration file "tsconfig.json". Please ensure it contains valid JSON and compiler options.
```

## Additional context

- Added 6 unit tests covering: missing config file, successful parsing, undefined parse result, error message content, undefined projectReferences, and config path forwarding.
- All tests pass.

### DIFF
--- a/lib/compiler/helpers/tsconfig-provider.ts
+++ b/lib/compiler/helpers/tsconfig-provider.ts
@@ -18,7 +18,12 @@ export class TsConfigProvider {
       undefined!,
       tsBinary.sys as unknown as ts.ParseConfigFileHost,
     );
-    const { options, fileNames, projectReferences } = parsedCmd!;
+    if (!parsedCmd) {
+      throw new Error(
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
Could you rebase to v12?

yogeshwaran-c:
Closing — both the fix and tests are already merged on v12.0.0 via #3332 and #3334.

PR #3357, opened 2026-04-09
fix(compiler): normalize output filename path separators

yogeshwaran-c:

Bug fix

Both `rspack-defaults.ts` and `webpack-defaults.ts` use `path.join()` to construct `output.filename`. On Windows, `join()` produces backslashes (e.g., `apps\my-app\main.js`), but webpack and rspack require forward slashes in `output.filename` regardless of platform.

## What is the new behavior?

Added `.replace(/\/g, '/')` to normalize backslashes to forward slashes in both files, matching the pattern already used elsewhere in the codebase (e.g., `swc-defaults.ts` has a `convertPath()` helper for this purpose).

### DIFF
--- a/lib/compiler/defaults/rspack-defaults.ts
+++ b/lib/compiler/defaults/rspack-defaults.ts
@@ -46,7 +46,10 @@ export const rspackDefaultsFactory = (
     devtool: isDebugEnabled ? 'inline-source-map' : false,
     target: 'node',
     output: {
-      filename: join(relativeSourceRoot, `${entryFilename}.js`),
+      filename: join(relativeSourceRoot, `${entryFilename}.js`).replace(
+        /\\/g,
+        '/',
+      ),
       ...(isEsm && {
         module: true,
         library: { type: 'module' },
--- a/lib/compiler/defaults/webpack-defaults.ts
+++ b/lib/compiler/defaults/webpack-defaults.ts
@@ -47,7 +47,10 @@ export const webpackDefaultsFactory = (
     devtool: isDebugEnabled ? 'inline-source-map' : false,
     target: 'node',
     output: {
-      filename: join(relativeSourceRoot, `${entryFilename}.js`),
+      filename: join(relativeSourceRoot, `${entryFilename}.js`).replace(
+        /\\/g,
+        '/',
+      ),
     },
     ignoreWarnings: [/^(?!CriticalDependenciesWarning$)/],
     externals: [externals() as any],


kamilmysliwiec:
Could you share a reproduction repository?

yogeshwaran-c:
Here's a minimal reproduction that demonstrates the bug without needing a full repository:

## Reproduction steps

1. On Windows, create a monorepo with a sub-app:
   \`\`\`bash
   mkdir my-workspace && cd my-workspace
   nest new my-workspace
   nest generate app my-api
   \`\`\`

2. Configure rspack (or webpack) as the builder in \`nest-cli.json\`:
   \`\`\`json
   {
     "compilerOptions": {
       "builder": "rspack"
     }
   }
   \`\`\`

3. Build the sub-app:
   \`\`\`bash
   nest build my-api
   \`\`\`

4. On Windows, the build produces: \`dist\apps\my-api\main.js\` (backslashes)
   On macOS/Linux, the build produces: \`dist/apps/my-api/main.js\` (forward slashes)

## The issue

In \`lib/compiler/defaults/rspack-defaults.ts\` (line 49) and \`lib/compiler/defaults/webpack-defaults.ts\` (line 50):

\`\`\`typescript
output: {
  filename: join(relativeSourceRoot, \`\${entryFilename}.js\`),
}
\`\`\`

\`path.join()\` uses the OS-native separator. On Windows, \`join('apps/my-api', 'main.js')\` returns \`'apps\my-api\main.js'\`.

## Evidence from test failures

The existing test \`test/lib/compiler/rspack/rspack-defaults.spec.ts:306\` already fails on Windows without this fix:

\`\`\`
× should set output filename based on relativeSourceRoot and entryFilename
  Expected: "apps/my-app"
  Received: "apps\my-app\main.js"
\`\`\`

This test was recently updated (#3334 / 8e5064f5) specifically because of the Windows path separator issue. The rspack-defaults test suite currently fails on Windows CI/local machines without this fix applied.

## Webpack-specific context

The webpack docs state that [\`output.filename\`](https://webpack.js.org/configuration/output/#outputfilename) uses forward slashes as the path separator regardless of platform:

> **Note that this option is called filename but you are still allowed to use something like \`"js/[name]/bundle.js"\` to create a folder structure.**

Webpack internally parses the filename as a path, and backslashes cause incorrect folder structures on Windows.

## The fix

Replacing backslashes with forward slashes on the filename ensures cross-platform consistency:

\`\`\`typescript
filename: join(relativeSourceRoot, \`\${entryFilename}.js\`).replace(/\\/g, '/'),
\`\`\`

This matches the pattern already used in \`lib/compiler/defaults/swc-defaults.ts\`, which has a \`convertPath()\` helper for the same purpose.

PR #3362, opened 2026-04-09
fix(build): warn when library assets are silently ignored

yogeshwaran-c:

Bug fix / Improvement

When building an application in a monorepo, assets configured for library projects in `nest-cli.json` are silently ignored. Users expect their library assets to be copied during the app build, but nothing happens and no feedback is given.

Ref: #1845

## What is the new behavior?

During the app build, if any library project in the monorepo has `compilerOptions.assets` configured, a warning is logged:

```
[INFO] Assets configured for library "my-lib" will not be copied during application build. Build the library separately or move assets to the application configuration.
```

This makes the behavior explicit and guides users toward a working solution.

## Additional context

The warning only fires for projects with `type: "library"` that have non-empty assets. It runs once per build per library, and doesn't affect the build process itself.

### DIFF
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -104,6 +104,7 @@ export class BuildAction extends AbstractAction {
         outDir,
         watchAssetsMode,
       );
+      this.warnOnIgnoredLibraryAssets(configuration, appName);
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
lgtm

Issue #3364, opened 2026-04-09
PluginMetadataGenerator generates dynamic imports without file extension with moduleResolution nodenext


VojtechVidra:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When using `swc` builder and `@nestjs/swagger` I need to generate metadata file and pass it manually with `loadPluginMetadata` to the SwaggerModule. The generated `metadata.ts` includes dynamic `import(./hello.dto)` statements which don't include file extension e.g. `hello.dto.js`, regardless of the `tsconfig.json` `moduleResolution` setting.

This doesn't actually break the application, the generated metadata file can still be used by SwaggerModule. However when the project is typechecked or when the `generate-metadata.ts` file is executed I get type errors.

### Minimum reproduction code

https://github.com/VojtechVidra/nestjs-cli-metadata-import-issue

### Steps to reproduce

1. `npm i`
2. `npm run generate-metadata`

### Expected behavior

When using `"moduleResolution": "nodenext"` the generated metadata file should include file extensions in the dynamic import statements.

### Package version

11.0.19

### NestJS version

11.1.18

### Node.js version

24.14.1

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

VojtechVidra:
I just realized if the file extensions are added manually the nestjs application won't work. Maybe the solution is to just add `// @ts-nocheck` at the top of the generated metadata file, next to the `/* eslint-disable */` 😄 .

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest-cli/pull/3404

PR #3365, opened 2026-04-10
fix(plugins): add ts-nocheck to generated metadata file

yogeshwaran-c:

Bug fix

The generated `metadata.ts` file contains dynamic `import()` statements without file extensions (e.g., `import("./hello.dto")`). When using `moduleResolution: nodenext`, TypeScript reports type errors because it requires explicit `.js` extensions in imports.

The imports work correctly at runtime — the issue is only with type-checking.

Ref: #3364

## What is the new behavior?

Added `// @ts-nocheck` directive to the generated metadata file, alongside the existing `/* eslint-disable */` comment. Since this file is entirely auto-generated and not meant to be manually edited or type-checked, suppressing TypeScript errors is the correct approach.

Generated file header is now:
```typescript
/* eslint-disable */
// @ts-nocheck
export default async () => { ... }
```

## Additional context

As noted by the reporter, adding `.js` extensions to the dynamic imports would break the NestJS application at runtime. `// @ts-nocheck` is the simplest correct fix that doesn't affect runtime behavior.

### DIFF
--- a/lib/compiler/plugins/plugin-metadata-printer.ts
+++ b/lib/compiler/plugins/plugin-metadata-printer.ts
@@ -74,10 +74,10 @@ export class PluginMetadataPrinter {
       options.outputDir!,
       options.filename ?? SERIALIZED_METADATA_FILENAME,
     );
-    const eslintPrefix = `/* eslint-disable */\n`;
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Instead of adding the ts-nocheck we should generate an ESM-compatible metadata file for ESM projects, and CJS for CJS

PR #3375, opened 2026-04-11
test(utils): add unit tests for is-module-available and runner factory

yogeshwaran-c:

Tests

Two utility modules have no test coverage on v12:
- `lib/utils/is-module-available.ts` — used by the build action to detect optional deps
- `lib/runners/runner.factory.ts` — factory for all runner implementations

### `is-module-available` (6 tests)
- Resolves a direct dependency (`typescript`)
- Resolves built-in Node.js modules (`fs`, `path`)
- Returns false for non-existent packages
- Returns false for invalid relative paths
- Returns false for an empty string

### `runner.factory` (6 tests)
- Creates the correct runner for each enum value (`SCHEMATIC`, `NPM`, `YARN`, `PNPM`)
- Throws for an unsupported runner value
- Returns a new instance on each call

## Test plan
- [x] All 12 tests pass

### DIFF
--- a/test/lib/runners/runner.factory.spec.ts
+++ b/test/lib/runners/runner.factory.spec.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest';
+import { NpmRunner } from '../../../lib/runners/npm.runner.js';
+import { PnpmRunner } from '../../../lib/runners/pnpm.runner.js';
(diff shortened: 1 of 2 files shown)


yogeshwaran-c:
Closing in favor of the consolidated PR #3388 as requested.

PR #3381, opened 2026-04-12
test(compiler): add unit tests for plugins loader

yogeshwaran-c:

Tests

No tests exist for `PluginsLoader` in `lib/compiler/plugins/plugins-loader.ts`.

## What is the new behavior?

Added 6 tests covering:
- Empty/undefined plugin list returns empty hooks
- Valid plugin loads before and after hooks
- Plugin without hooks throws descriptive error
- Missing plugin throws "not installed" error
- Object-format plugins with options are handled correctly

## Test plan
- [x] All 6 tests pass


yogeshwaran-c:
Closing in favor of the consolidated PR #3388 as requested.

PR #3399, opened 2026-04-17
feat(cli): add rspack config path option to build and start commands

yogeshwaran-c:

Feature (small). Closes a CLI gap — no existing issue.

`getRspackConfigPath` already honours `cmdOptions.rspackPath` (there's even a unit test for it: `test/lib/compiler/helpers/get-rspack-config-path.spec.ts > should return the CLI rspackPath option when provided`). However, neither the `build` nor the `start` command registers the flag with commander, so `options.rspackPath` is always `undefined` at runtime and projects using the rspack builder can only point at a custom rspack config via `nest-cli.json`.

This is inconsistent with webpack, which has had `--webpackPath` wired for a long time.

## What is the new behavior?

Register `--rspackPath [path]` on both the `build` and `start` commands, mirroring the existing `--webpackPath` option. The option is forwarded into the command contexts, which the action already passes to `getRspackConfigPath`.

Changed files:

- `commands/build.command.ts` — register `--rspackPath`, forward to `BuildCommandContext`
- `commands/start.command.ts` — register `--rspackPath`, forward to `StartCommandContext`
- `commands/context/build.context.ts` / `start.context.ts` — add `rspackPath?: string`
- `test/actions/build.action.spec.ts` — new test asserting the option reaches `getRspackConfigPath`

No existing behavior changes: when the flag is not passed, `options.rspackPath` stays `undefined` and the helper falls back to `builder.options.configPath` / default filename as before.

## Test plan

- [x] `npm run build` clean (tsc)
- [x] `npx vitest run test/actions/build.action.spec.ts` — 4/4 pass
- [x] `npx vitest run test/lib/compiler/helpers/get-rspack-config-path.spec.ts` — 8/8 pass
- [x] No change to unrelated tests locally (the 5 pre-existing Windows failures in `tsconfig-paths.hook.spec.ts` are tracked separately in #3398)

### DIFF
--- a/commands/build.command.ts
+++ b/commands/build.command.ts
@@ -35,6 +35,7 @@ export class BuildCommand extends AbstractCommand {
       )
       .option('--silent', 'Suppress informational compiler logs.')
       .option('--webpackPath [path]', 'Path to webpack configuration.')
+      .option('--rspackPath [path]', 'Path to rspack configuration.')
       .option('--tsc', 'Use typescript compiler for compilation.')
       .option(
         '--preserveWatchOutput',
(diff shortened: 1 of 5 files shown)


kamilmysliwiec:
lgtm

PR #3401, opened 2026-04-17
feat(compiler): rewrite path aliases in emitted declaration files

### LINKED ISSUE #1749: Support for relative path conversion on .d.ts files

If I use nest build and my project emits .d.ts files, they still have to use the aliased path on the imports, what not happen with .js.
 
For test purposes, let's have this path configured:
```
    "paths": {
      "@core/*": [
        "./src/*"
      ]
```
I have one file with this line:
```
export * from '@core/test';
```
When I transpile it emitting types, the .js looks fine:
```
__exportStar(require("./test"), exports);
```
Which is fine, but the .d.ts still looks like this:

```ts
export * from '@core/test';
```

This causes a bad type detection if I use nest build to compile a library, for example, and try to import it into some other project.

Replace the aliased path by the relative one also in the .d.ts files when compilng

_No response_

 If the aliased path could be replaced by the relative one also in the .d.ts files, it would allow us to use nest build even to compile a library, for example, letting us have the easiness of using the aliased path in any kind o project

yogeshwaran-c:

Bug-fix backport from master to v12.0.0.

On master, [PR #3318 / commit 73e1aeb4](https://github.com/nestjs/nest-cli/commit/73e1aeb4) (closes #1749) wired the `tsconfigPaths` transformer into `afterDeclarations` in both `Compiler.run` and `WatchCompiler.overrideCreateProgramFn`, so that path aliases such as `~/foo` in emitted `.d.ts` files get rewritten to relative paths, matching the `.js` output.

That change never reached `v12.0.0`: `git log upstream/v12.0.0 -- lib/compiler/compiler.ts` stops at `fix(compiler): run tsconfig paths plugin before user plugins in tsc compiler`, with no follow-up. On v12.0.0 the transformer is still only installed on `before`, so consumers importing a library's `.d.ts` see unresolved aliases like:

```ts
// dist/main.d.ts (before this PR)
export { Foo } from '~/foo';
export { Bar } from '~/bar';
```

## What is the new behavior?

`.d.ts` output now gets the same transformer treatment as the JS output:

```ts
// dist/main.d.ts (after this PR)
export { Foo } from './foo';
export { Bar } from './bar';
```

## Additional context

- Minimal port — identical two-line additions to `compiler.ts` and `watch-compiler.ts`.
- Master's test file pairs a new declaration-file describe block with a `describe.skip` for the flaky older tests. On v12.0.0 those older tests have already been replaced by comprehensive CJS/ESM suites (`test/lib/compiler/hooks/tsconfig-paths.hook.spec.ts`), so this PR keeps those intact and only appends the new `declaration files (afterDeclarations)` describe block plus its fixtures.
- The new `createSpecWithDeclarations` helper normalizes emit paths to forward slashes so the assertions work on Windows as well as CI.
- Closes [#1749](https://github.com/nestjs/nest-cli/issues/1749) for the v12 line.

## Test plan

- [x] `npm run build` clean (tsc)
- [x] The two new declaration-file tests pass locally
- [x] No change to the existing CJS/ESM describe blocks
- [x] Pre-existing Windows-only failures in the unchanged `createSpec` helper are tracked in #3398; this PR does not regress them

### DIFF
--- a/lib/compiler/compiler.ts
+++ b/lib/compiler/compiler.ts
@@ -62,7 +62,9 @@ export class Compiler extends BaseCompiler {
           ? [tsconfigPathsPlugin, ...before]
           : before,
         after,
-        afterDeclarations,
+        afterDeclarations: tsconfigPathsPlugin
+          ? [tsconfigPathsPlugin, ...afterDeclarations]
+          : afterDeclarations,
       },
     );
 
--- a/lib/compiler/watch-compiler.ts
+++ b/lib/compiler/watch-compiler.ts
@@ -147,6 +147,7 @@ export class WatchCompiler extends BaseCompiler<TypescriptWatchCompilerExtras> {
         );
         if (tsconfigPathsPlugin) {
           before.unshift(tsconfigPathsPlugin);
+          afterDeclarations.unshift(tsconfigPathsPlugin);
         }
 
         transforms.before = before.concat(transforms.before || []);
--- a/test/lib/compiler/hooks/fixtures/aliased-dts-imports/src/bar.ts
+++ b/test/lib/compiler/hooks/fixtures/aliased-dts-imports/src/bar.ts
@@ -1,0 +1,3 @@
+export class Bar {
+  value: number = 0;
+}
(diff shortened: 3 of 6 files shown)


kamilmysliwiec:
just merged master so it's already there

PR #3445, opened 2026-05-16
test(compiler): cover Compiler.run emit and diagnostic exit paths

yogeshwaran-c:
## Summary

Adds unit tests for the tsc-based `Compiler.run` (no existing direct coverage).

The new spec verifies:
- `onSuccess` is invoked and `program.emit()` runs when there are no diagnostics.
- The compiler falls back to `createProgram` when `createIncrementalProgram` is unavailable, and uses `createIncrementalProgram` when the loaded TypeScript exposes it.
- `process.exit(1)` is called and diagnostics are printed when pre-emit or emit diagnostics are present.
- A clean run completes when `onSuccess` is not supplied.

TypeScript, the `TsConfigProvider`, the `TypeScriptBinaryLoader`, and the `PluginsLoader` are stubbed at the class boundary so the test stays focused on `Compiler` orchestration logic.

## Test plan
- [x] `npx vitest run test/lib/compiler/compiler.spec.ts` — 6 passed

### DIFF
--- a/test/lib/compiler/compiler.spec.ts
+++ b/test/lib/compiler/compiler.spec.ts
@@ -1,0 +1,197 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { Compiler } from '../../../lib/compiler/compiler.js';
+import { TsConfigProvider } from '../../../lib/compiler/helpers/tsconfig-provider.js';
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
moving forward, could you create a single PR with all the test updates instead of X PRs? 🙏

Issue #3466, opened 2026-06-11
Cannot run 'nest start' on a path containing '-no-'


nicolas-spott:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`nest start` crashes when the path contains a `no` segment:

```
TypeError: Cannot read properties of undefined (reading 'toUpperCase')
    at eval (file:///home/projects/nestjs-typescript-starter-snkda8se/broken-dir-no-broken/node_modules/.pnpm/@nestjs+cli@11.0.23_@swc+cli@0.6.0_@swc+core@1.15.41_@types+node@22.19.21_prettier@3.8.4/node_modules/@nestjs/cli/lib/utils/remaining-flags.js#cjs:34:30)
    at Array.reduce (<anonymous>)
    at camelCase (file:///home/projects/nestjs-typescript-starter-snkda8se/broken-dir-no-broken/node_modules/.pnpm/@nestjs+cli@11.0.23_@swc+cli@0.6.0_@swc+core@1.15.41_@types+node@22.19.21_prettier@3.8.4/node_modules/@nestjs/cli/lib/utils/remaining-flags.js#cjs:33:28)
    at eval (file:///home/projects/nestjs-typescript-starter-snkda8se/broken-dir-no-broken/node_modules/.pnpm/@nestjs+cli@11.0.23_@swc+cli@0.6.0_@swc+core@1.15.41_@types+node@22.19.21_prettier@3.8.4/node_modules/@nestjs/cli/lib/utils/remaining-flags.js#cjs:17:33)
    at Array.filter (<anonymous>)
    at getRemainingFlags (file:///home/projects/nestjs-typescript-starter-snkda8se/broken-dir-no-broken/node_modules/.pnpm/@nestjs+cli@11.0.23_@swc+cli@0.6.0_@swc+core@1.15.41_@types+node@22.19.21_prettier@3.8.4/node_modules/@nestjs/cli/lib/utils/remaining-flags.js#cjs:8:10)
    at Command.eval (file:///home/projects/nestjs-typescript-starter-snkda8se/broken-dir-no-broken/node_modules/.pnpm/@nestjs+cli@11.0.23_@swc+cli@0.6.0_@swc+core@1.15.41_@types+node@22.19.21_prettier@3.8.4/node_modules/@nestjs/cli/commands/start.command.js#cjs:94:67)
[...]

PR #3467, opened 2026-06-12
fix(cli): fix treating path args as flags

QusaiAlbonni:
## PR Checklist

## PR Type

- [x] Bugfix

Issue Number: #3466

## What is the new behavior?
`getRemainingFlags` now returns an empty array for when no flags exist, also when values are checked for if they already exist as an option in commander it will now only check flags rather than everything which falsly checks paths and values leading to errors like the one in the issue

- [x] No


kamilmysliwiec:
can you create a second PR with the same set of changes targeting v12 branch as well? https://github.com/nestjs/nest-cli/pull/3280 ideally, we would have an e2e test for this in that branch too

QusaiAlbonni:
#3468

PR #3468, opened 2026-06-13
feat(cli): get remaining flags now returns empty when no flag is found

QusaiAlbonni:
## PR Checklist

## PR Type

- [x] Feature

Issue Number: #3466

## What is the new behavior?
`getRemainingFlags` now returns an empty array when no flags are found instead of returning all arguments.

In v11/master this behavior contributes to a downstream failure and is therefore treated as a bug fix. However, in the v12 branch, #3355 modified camelCase to handle these values safely, so the incorrect return value and processing of it no longer causes a user-visible failure. As a result, this change is classified as an enhancement/semantic correction rather than a bug fix as it might prevent future bugs from emerging.

- [x] No


kamilmysliwiec:
could you add an e2e test?

QusaiAlbonni:
> could you add an e2e test?

just to be clear an e2e test for the case of -no- in argv[0] just like the original issue? if so yes I can.

kamilmysliwiec:
an e2e test which confirms that if you run a command with a path containing -no- it doesnt fail

QusaiAlbonni:
> an e2e test which confirms that if you run a command with a path containing -no- it doesnt fail

`localBinaries` is importing commands from the project local node_modules which is why the e2e test is failing in circleci as the fix was not applied yet to the current release in v11

PR #3469, opened 2026-06-16
fix(runners): remove duplicated binary in failed command message

iamkanguk97:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When a command executed through `AbstractRunner` fails, the CLI prints the failed command with the package manager binary duplicated. For example, during `nest new` with pnpm:

```
Failed to execute command: pnpm pnpm install --strict-peer-dependencies=false --reporter=silent
```

In `run()`, the inner `command` variable already contains the binary (`[this.binary, ...this.args, ...args].join(' ')`), but the failure message prepended `this.binary` once more:

```ts
console.error(red(MESSAGES.RUNNER_EXECUTION_ERROR(`${this.binary} ${command}`)));
```

This produces the duplicated `pnpm pnpm`.
The bug is package-manager agnostic (npm/yarn would print the same).

Issue Number: nestjs/nest#17135

## What is the new behavior?

The failure message now prints the command without duplicating the binary:

```
Failed to execute command: pnpm install --strict-peer-dependencies=false --reporter=silent
```

The inner variable is renamed to `fullCommand` to remove the shadowing of the outer `command` parameter, and the error message reuses it instead of prepending the binary again. A regression test for `AbstractRunner.run()` was added.

```
[ ] Yes
[x] No
```

## Other information

Verified locally: the new test fails against the previous code and passes with the fix, and the full test suite (`npm test`) stays green.

### DIFF
--- a/lib/runners/abstract.runner.ts
+++ b/lib/runners/abstract.runner.ts
@@ -22,8 +22,8 @@ export class AbstractRunner {
       shell: true,
     };
     return new Promise<null | string>((resolve, reject) => {
-      const command = [this.binary, ...this.args, ...args].join(' ');
-      const child: ChildProcess = spawn(command, options);
+      const fullCommand = [this.binary, ...this.args, ...args].join(' ');
+      const child: ChildProcess = spawn(fullCommand, options);
       if (collect) {
         child.stdout!.on('data', (data) =>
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #3478, opened 2026-07-10
fix(compiler): fail fast when TypeScript lacks the programmatic API (TS 7)

### LINKED ISSUE #3477: nest start crashes with "getParsedCommandLineOfConfigFile is not a function" on TypeScript 7.0.2

### Current behavior

`nest start` / `nest start --watch` crashes immediately with:

```
Error  tsBinary.getParsedCommandLineOfConfigFile is not a function
```

Root cause: `lib/compiler/helpers/tsconfig-provider.js` calls `tsBinary.getParsedCommandLineOfConfigFile(configPath, undefined, tsBinary.sys)`, but `typescript@7.0.2`'s CommonJS module does not export that function at all:

```
node -e "const ts = require('typescript'); console.log(typeof ts.getParsedCommandLineOfConfigFile)"
// -> undefined
```

This is different from #3332 / #3345 (which handle the function *returning* `undefined`) — here the function itself doesn't exist on the module, so it throws a `TypeError` before any null-check could run.

This is expected per Microsoft's own TypeScript 7.0 migration notes: the stable programmatic/compiler API is not shipping until TypeScript 7.1 — TS 7.0 only guarantees the `tsc` CLI, not the library surface tools like `@nestjs/cli` depend on.

### Minimum reproduction code

No separate repo — 3-line reproduction below is enough to hit it on any existing NestJS project.

### Steps to reproduce

1. `nest new repro`
2. `npm i typescript@7.0.2`
3. `npm run start`
4. See error above

### Expected behavior

Either `nest start`/`nest build` work against TypeScript 7, or the CLI detects the incompatibility and fails with a clear, actionable error (e.g. "TypeScript 7's programmatic API is not yet supported, please use TypeScript 6 until 7.1") instead of a raw `TypeError`.

### Package version

11.0.23

### NestJS version

11.1.19

### Node.js version

v22.18.0

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

Plain `tsc --noEmit` / `tsc --noEmit -p tsconfig.json` is unaffected — only tools that `require('typescript')` and call the programmatic API break.

ruban-s:
## PR Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added
- [x] Docs have been added / updated (n/a — internal behavior + error message)

`nest build` / `nest start --watch` crash immediately with a raw `TypeError` when `typescript@7.x` is installed:

```
Error  tsBinary.getParsedCommandLineOfConfigFile is not a function
```

Root cause: TypeScript 7.0 ships the native (Go) `tsc` executable only — its Node module does **not** expose the programmatic compiler API. Per Microsoft's migration notes, that API is expected to return in TypeScript 7.1. The CLI relies on it, so it throws deep in the build with no indication of the real problem:

```js
const ts = require('typescript');
ts.version;                                 // "7.0.2"
typeof ts.getParsedCommandLineOfConfigFile; // "undefined"
typeof ts.createProgram;                    // "undefined"
```

This affects **both** builders — the default `tsc` path (`helpers/tsconfig-provider.ts`) and the SWC path (`swc/type-checker-host.ts` calls `createProgram` / `createWatchProgram` / `getPreEmitDiagnostics`, all `undefined`), so SWC is not a workaround.

Closes #3477.

## What is the new behavior?

`TypeScriptBinaryLoader.load()` — the single choke point every consumer goes through — now asserts the programmatic API is present right after resolving `typescript`, and throws an actionable error otherwise:

```
The installed TypeScript version (7.0.2) does not expose the programmatic compiler API
that the Nest CLI requires. TypeScript 7.0 ships the "tsc" executable only; the compiler
API is expected to return in 7.1. Please install TypeScript 6 (e.g. "npm i -D typescript@^6")
until then.
```

Because the guard lives in the loader, it covers every entry point (default builder, SWC type-checker host, plugin metadata generator) with one small change.

Note: I used **feature detection** (`typeof getParsedCommandLineOfConfigFile !== 'function'`) rather than a hard `major === 7` version gate on purpose — when TypeScript 7.1 restores the API, the CLI starts working again automatically with no follow-up change needed. The installed version is still surfaced in the message for the user's benefit.

Plain `tsc --noEmit` remains unaffected (only tools that `require('typescript')` and call the API break).

- [ ] Yes
- [x] No

### DIFF
--- a/lib/compiler/typescript-loader.ts
+++ b/lib/compiler/typescript-loader.ts
@@ -1,4 +1,5 @@
 import * as ts from 'typescript';
+import { CLI_ERRORS } from '../ui';
 
 export class TypeScriptBinaryLoader {
   private tsBinary?: typeof ts;
@@ -8,17 +9,28 @@ export class TypeScriptBinaryLoader {
       return this.tsBinary;
     }
 
+    let tsBinary: typeof ts;
     try {
       const tsBinaryPath = require.resolve('typescript', {
         paths: [process.cwd(), ...this.getModulePaths()],
       });
       // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const tsBinary = require(tsBinaryPath);
-      this.tsBinary = tsBinary;
-      return tsBinary;
+      tsBinary = require(tsBinaryPath);
     } catch {
       throw new Error(
         'TypeScript could not be found! Please, install "typescript" package.',
+      );
+    }
+
+    this.assertProgrammaticApiIsSupported(tsBinary);
+    this.tsBinary = tsBinary;
+    return tsBinary;
+  }
+
+  private assertProgrammaticApiIsSupported(tsBinary: typeof ts): void {
+    if (typeof tsBinary.getParsedCommandLineOfConfigFile !== 'function') {
+      throw new Error(
+        CLI_ERRORS.UNSUPPORTED_TYPESCRIPT_VERSION(tsBinary.version),
       );
     }
   }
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
lgtm

PR #3521, opened 2026-08-24
refactor(compiler): replace glob with an internal fs-based helper

### LINKED ISSUE #3520: Replace glob with native runtime implementations

At the moment, we rely on `glob` library in [asset-manager.ts](https://github.com/nestjs/nest-cli/blob/533d8c484018758e1a734b0580dacbfd23aba6ff/lib/compiler/assets-manager.ts#L4). It's a third party we may no need given the usage we have in the CLI. There are also compelling alternatives listed in https://e18e.dev/docs/replacements/glob

Since Node.js 22, a similar [glob](https://nodejs.org/api/fs.html#fspromisesglobpattern-options) implementation also exists. This would save a few Mb when downloading the CLI.

Supported engine also specifies Node 22 which is EOL. It could be opportunity to review it at the same time.

_No response_

My main motivation is to reduce disk space and reduce bandwidth when we can by removing dependencies when we can. Here's a page with more details https://e18e.dev/learn/cleanup.html

micalevisk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`glob` is a runtime dependency used in exactly one place — `lib/compiler/assets-manager.ts`, at three call sites — for a narrow slice of its feature set: one pattern, one optional `ignore` pattern, and `dot`.

It pulls in `path-scurry`, `lru-cache` and `minipass` as exclusive transitives, roughly 3.4 MB installed, for that one usage.

Issue Number: closes #3520

## What is the new behavior?

`glob` is removed and replaced by a small internal helper (`lib/compiler/helpers/glob.ts`) built on `fs.readdirSync` and `minimatch`. `minimatch` was already a direct dependency, used by `lib/compiler/swc/swc-compiler.ts`, so nothing new is added.

**Why not `fs.globSync`?** The issue suggests Node's built-in, but it has no `dot` option and silently ignores one if passed. `dot: true` is load-bearing here — it was added in 3664ed06 to unify watched and non-watched asset copying, so dropping it would stop dotfile assets being copied. `fs.globSync` also requires Node 22, while `engines.node` is `>= 20.11`, so it would have forced an engine bump regardless.

The helper is a drop-in for `glob`'s `sync` export. The non-obvious behaviours it preserves, each pinned by a test:

- **A trailing globstar matches zero segments** — `a/**` matches `a` itself. `minimatch` alone does not do this.
- **`ignore` prunes the ignored directory itself**, not only its contents.
- **A globstar does not recurse *through* a symlinked directory**, but the explicit segments after it keep matching normally — `assets/**` stops at the symlink, `assets/**/*` reaches one level in. This is derived as a depth budget from the pattern, which also makes symlink cycles terminate.
- **An unreadable subdirectory is skipped**, not fatal — one `EACCES` deep in an asset tree must not silently empty the result.
- **Only real glob metacharacters split the pattern.** An earlier revision of this branch also treated `@ ( ) [ ] { } ! +` as metacharacters, which collapsed the scan base to a distant ancestor on any project path containing them — `user@company.com`, `Program Files (x86)`, `node_modules/@scope/` — producing a **silent zero-asset build**. There are regression tests for both shapes.

Parity was verified differentially against `glob@13` on fixtures containing `@` and `(x86)` in the path, a symlink cycle, and an unreadable subdirectory.

```
[ ] Yes
[x] No
```

Asset copying is intended to behave exactly as before. No public API changes, and no change to `engines.node`.

## Other information

**Docs checkbox** is unchecked because this is an internal refactor with no user-facing surface — there is nothing in the docs describing how asset globbing is implemented. Happy to add something if you disagree.

**A weak existing test, worth flagging.** `test/e2e/library-assets.e2e-spec.ts` configures `assets: ['templates/**/*.hbs']` but only asserts that `main.js` files exist — it never asserts the `.hbs` file reached `dist`, so it passes even when zero assets are copied. It did not catch the metacharacter bug described above. Happy to tighten it here or in a follow-up.

**Lockfile note.** Regenerating `package-lock.json` also drops six `@module-federation/*` entries. These are unrelated to this change — an optional-peer closure npm stopped materialising after the `@rspack/core` v1 → v2 bump in 54d429df, which `npm install` drops deterministically on the current lockfile regardless of this PR. Restoring them by hand would produce a lockfile npm did not generate. Happy to split them out if you'd prefer.

**Follow-ups, deliberately not included here:**

- `assets-manager.ts` passes `dot: true` at two call sites but not at the directory-expansion one, so a dotfile inside a matched directory is copied in watch mode but not in a plain build. Preserved verbatim to keep this change behaviour-neutral, but it looks like a genuine latent bug and deserves its own issue.
- `globEntriesSync` exposes entry types, so the per-path `statSync` calls in `assets-manager` could be dropped. Left out to keep this diff reviewable. Note for whoever picks it up: that helper's `isFile`/`isDirectory` flags follow symlinks in the literal-path branch but not in the walk branch, so a naive swap would stop copying symlinked asset files.

---

Thanks to Claude Opus 5

### DIFF
--- a/lib/compiler/assets-manager.ts
+++ b/lib/compiler/assets-manager.ts
@@ -1,6 +1,5 @@
 import * as chokidar from 'chokidar';
 import { copyFileSync, mkdirSync, rmSync, statSync } from 'fs';
-import { sync } from 'glob';
 import { dirname, join, sep } from 'path';
 import {
   ActionOnFile,
@@ -10,6 +9,7 @@ import {
 } from '../configuration/index.js';
 import { ERROR_PREFIX } from '../ui/index.js';
 import { copyPathResolve } from './helpers/copy-path-resolve.js';
+import { globSync } from './helpers/glob.js';
 import { getValueOrDefault } from './helpers/get-value-or-default.js';
 import {
   areOutsidePathsAllowed,
@@ -162,7 +162,7 @@ export class AssetsManager {
         };
 
         if (isWatchEnabled || item.watchAssets) {
-          const matchedPaths = sync(item.glob, {
+          const matchedPaths = globSync(item.glob, {
             ignore: item.exclude,
             dot: true,
           });
@@ -227,7 +227,7 @@ export class AssetsManager {
           this.watchers.push(watcher);
           this.watcherReadyPromises.push(settled);
         } else {
-          const matchedPaths = sync(item.glob, {
+          const matchedPaths = globSync(item.glob, {
             ignore: item.exclude,
             dot: true,
           });
@@ -235,7 +235,7 @@ export class AssetsManager {
             ? matchedPaths.filter((matched) => statSync(matched).isFile())
             : matchedPaths.flatMap((matched) => {
                 if (statSync(matched).isDirectory()) {
-                  return sync(`${matched}/**/*`, {
+                  return globSync(`${matched}/**/*`, {
                     ignore: item.exclude,
                   }).filter((file) => statSync(file).isFile());
                 }
--- a/lib/compiler/assets-manager.ts
+++ b/lib/compiler/assets-manager.ts
@@ -273,8 +273,7 @@ export class AssetsManager {
       }
 
       const libAssets = libProject.compilerOptions?.assets as
-        | Asset[]
-        | undefined;
+        Asset[] | undefined;
       if (!libAssets || libAssets.length <= 0) {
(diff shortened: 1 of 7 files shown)


kamilmysliwiec:
excellent, thank you!

Issue #2687, opened 2024-08-08
after nest build the nest-cli defined assets are only placed into the expected outdir directories if watchAssets is true


andreasvh-conceto:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

After `npm run build` the files are not placed in the expected directories. Only if `watchAssets` is set to true the files are copied to their desired outdirs.

### Minimum reproduction code

https://github.com/andreasvh-conceto/nestcli-reproduce-asset-include

### Steps to reproduce

1. `git clone git@github.com:andreasvh-conceto/nestcli-reproduce-asset-include.git`
2. `cd nestcli-reproduce-asset-include`
3. `npm i`
4. Look inside the `nest-cli.json` to check expectations
5. `npm run build` and look inside the dist folder
![image](https://github.com/user-attachments/assets/d617c678-f67f-4134-a00c-1db63f6dbd62)
6. set `watchAssets: true` in the `nest-cli.json`
7. `npm run build`
8. Look again in the dist directory and compare it to step 5
![image](https://github.com/user-attachments/assets/3b5a3d09-dd7f-43d3-8fb1-ebba7ba84ff3)

### Expected behavior

If `watchAssets` is not set or is set to true/false will not have an effect on the outdir directory structure after an `npm run build`. The files/directories defined in `nest-cli.json` will be copied as expected to their outdirs.

### Package version

10.4.2

### NestJS version

10.3.10

### Node.js version

v20.12.1

### In which operating systems have you tested?

- [X] macOS
- [X] Windows
- [X] Linux

### Other

Could be related to: https://github.com/nestjs/nest-cli/issues/2677

Hardware Overview:

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro17,1
  Model Number:	Z11C000JAD/A
  Chip:	Apple M1
  Total Number of Cores:	8 (4 performance and 4 efficiency)
  Memory:	16 GB
  System Firmware Version:	10151.121.1
  OS Loader Version:	10151.121.1
  Serial Number (system):	FVFFF2NPQ05P
  Hardware UUID:	F8CA8278-2342-568C-AF29-5A89497A28A4
  Provisioning UDID:	00008103-000E2098143A001E
  Activation Lock Status:	Disabled

kamilmysliwiec:
Why if you change this

```json
{"include": "../certs/*", "outDir": "./dist/certs1"},
{"include": "../certs", "outDir": "./dist/certs2"}
```

[...]

Issue #2653, opened 2024-07-17
Hidden assets are not copied


addun:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Potential Commit/PR that introduced the regression

e19c2d3b58aed6d38e0e2109e5a378a25256b9d1

### Versions

9 -> 10

### Describe the regression

Assets are not copied when files start with `.`

### Minimum reproduction code

1. Create a file in `app/config/envs/.development`
3. Add configuration to copy this file
```json
  "compilerOptions": {
    "assets": ["config/envs/*"]
  }
```
3. Build your application

https://stackblitz.com/edit/nestjs-typescript-starter-dhma8x?file=README.md

### Expected behavior

In v9 (@nestjs/cli@9) all files inside `config/envs/*` were copied to `/dist` file. After migration to v10 they didn't. To fix this problem you need to explicitly add `.` prefix.

```json
  "compilerOptions": {
    "assets": ["config/envs/.*"] // with `.` is works
  }
```

### Other

Glob by default do NOT copy hidden files https://stackoverflow.com/questions/77509905/the-glob-npm-package-does-not-matches-the-file-which-name-begins-from-dot

kamilmysliwiec:
We could expose a configuration attribute to control that behavior - would you like to create a PR for this?

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest-cli/pull/2660

PR #3523, opened 2026-08-24
fix(compiler): copy dotfiles when expanding a directory asset

### LINKED ISSUE #3522: Assets: dotfiles inside a directory asset are copied by `--watchAssets` but not by a plain `nest build`

### Current behavior

When an `assets` entry is a **plain directory path (no trailing `*`)**, dotfiles inside that directory are copied in `--watchAssets` mode but silently skipped by a plain `nest build`.

Given `"assets": ["config"]` and `src/config/envs/.development`:

| Command | `dist/config/envs` contains |
| --- | --- |
| `nest build` | `prod.env` only — **`.development` is missing** |
| `nest build --watchAssets` | `.development`, `prod.env` |

Image

The cause is in `lib/compiler/assets-manager.ts`. `dot: true` is passed at two of the three `glob` call sites, but not at the directory-expansion one:

```ts
// L165 — watch branch: has dot
const matchedPaths = sync(item.glob, { ignore: item.exclude, dot: true });

// L230 — plain-build branch: has dot
const matchedPaths = sync(item.glob, { ignore: item.exclude, dot: true });

const files = item.glob.endsWith('*')
  ? matchedPaths.filter((matched) => statSync(matched).isFile())
  : matchedPaths.flatMap((matched) => {
      if (statSync(matched).isDirectory()) {
        // L238 — directory expansion: NO dot  <-- bug
        return sync(`${matched}/**/*`, {
          ignore: item.exclude,
        }).filter((file) => statSync(file).isFile());
      }
      return matched;
    });
```

Because `glob` does not match dotfiles unless `dot: true` is set, the expansion at L238 drops every dotfile under the matched directory. Chokidar, on the watch side, has no such default, so the same config copies the file in watch mode.

This is the exact behavior #2660 set out to unify ("`chokidar.watch` listen also for hidden files but `glob` do not"), and it is a resurfacing of #2653.

### Minimum reproduction code

No repo needed — this reproduces from a 5-file project; full script under "Steps to reproduce".

### Steps to reproduce

```bash
mkdir repro && cd repro

cat > package.json <<'EOF'
{ "name": "repro", "version": "1.0.0", "private": true }
EOF

cat > tsconfig.json <<'EOF'
{ "compilerOptions": { "module": "commonjs", "target": "ES2021", "rootDir": "./src", "outDir": "./dist" } }
EOF

cat > nest-cli.json <<'EOF'
{ "collection": "@nestjs/schematics", "sourceRoot": "src", "compilerOptions": { "assets": ["config"] } }
EOF

mkdir -p src/config/envs
echo "console.log('hi');" > src/main.ts
printf 'SECRET=1' > src/config/envs/.development
printf 'X=1'      > src/config/envs/prod.env

npm i -D @nestjs/cli@11.0.24
npm i -D "typescript@^6"   # the CLI does not support TS 7 yet

# 1) plain build -> .development is MISSING
rm -rf dist && npx nest build && find dist

# 2) watch-assets build -> .development IS copied.
#    This one does not exit on its own: wait a few seconds, Ctrl-C, then inspect.
rm -rf dist && npx nest build --watchAssets
find dist
```

Observed:

```
$ rm -rf dist && npx nest build && find dist
dist/config/envs/prod.env
dist/main.js

$ rm -rf dist && npx nest build --watchAssets   # Ctrl-C after a few seconds
$ find dist
dist/config/envs/.development     <-- present only here
dist/config/envs/prod.env
dist/main.js
```

Note that the wildcard form (`"assets": ["config/envs/*"]`) is **not** affected — it takes the `endsWith('*')` branch, which does receive `dot: true`. Only entries that do not end in `*` and resolve to a directory hit the bug.

### Expected behavior

`nest build` and `nest build --watchAssets` should copy the same set of files. A dotfile inside a directory asset should be copied in both modes, consistent with the intent of #2660.

Concretely, passing `dot: true` at the third call site as well:

```ts
return sync(`${matched}/**/*`, {
  ignore: item.exclude,
  dot: true,
}).filter((file) => statSync(file).isFile());
```

### Package version

11.0.24

### NestJS version

N/A

### Node.js version

26.3.0

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

**How the fix from #2660 got lost.** Both PRs branched from the same blob of `assets-manager.ts` (`c71bef10`), and neither is an ancestor of the other:

| Date | Commit | Event |
| --- | --- | --- |
| 2024-07-19 | `3664ed06` | #2660 authored. At that time the else-branch had a **single** `sync()` call; the PR adds `dot: true` to it. |
| 2024-08-09 | `793f3193` | #2687 merged (*"copy nested files and directories when no wildcard"*). It **replaces** that one call with two — outer `matchedPaths` + inner <code>sync(\`${matched}/**/*\`) </code>. Its base predates #2660, so neither call has `dot`. |
| 2024-10-21 | `486ac60e` | Merge commit *"chore: resolve conflicts"* finally lands #2660. The resolution puts `dot: true` back on the **outer** call, but the inner expansion call — which #2660's author never saw — is left without it. |

So this is not something #2660 overlooked; it is a merge-resolution gap against a call site introduced while #2660 was open.

**Why #2653 looks fixed.** #2653's own reproduction used `config/envs/*`, which takes the `endsWith('*')` path and genuinely works today. Only the bare-directory form still fails, which is why this went unnoticed.

**Affected versions.** Present in `@nestjs/cli@11.0.24` (current `latest`, verified above) and on `master` at `12.0.0-alpha.5`. Verified independently against `glob@13.0.6` — the version `master` depends on — that `sync('<dir>/**/*')` without `dot` omits dotfiles while the same call with `dot: true` includes them.

**Note on scope.** The fix copies strictly more files than today, so it is technically an observable change — but it restores the `build` / `--watchAssets` parity that #2660 established and matches what the same config already does in watch mode.

micalevisk:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: closes #3522

When an `assets` entry is a **plain directory path (no trailing `*`)**, dotfiles inside that directory are copied in `--watchAssets` mode but silently skipped by a plain `nest build`.

Given `"assets": ["config"]` and `src/config/envs/.development`:

| Command | `dist/config/envs` contains |
| --- | --- |
| `nest build` | `prod.env` only — **`.development` is missing** |
| `nest build --watchAssets` | `.development`, `prod.env` |

`lib/compiler/assets-manager.ts` has three glob call sites, and only two of them pass `dot: true`:

```ts
// watch branch: has dot
const matchedPaths = globSync(item.glob, { ignore: item.exclude, dot: true });

// plain-build branch: has dot
const matchedPaths = globSync(item.glob, { ignore: item.exclude, dot: true });

const files = item.glob.endsWith('*')
  ? matchedPaths.filter((matched) => statSync(matched).isFile())
  : matchedPaths.flatMap((matched) => {
      if (statSync(matched).isDirectory()) {
        // directory expansion: NO dot  <-- bug
        return globSync(`${matched}/**/*`, {
          ignore: item.exclude,
        }).filter((file) => statSync(file).isFile());
      }
      return matched;
    });
```

Since glob does not match dotfiles unless `dot: true` is set, the expansion drops every dotfile under the matched directory. Chokidar, on the watch side, has no such default — hence the divergence.

The wildcard form (`"assets": ["config/envs/*"]`) is **not** affected: it takes the `endsWith('*')` branch, which does receive `dot: true`. Only entries that do not end in `*` and resolve to a directory hit this.

**How the fix from #2660 got lost.** Both PRs branched from the same blob of `assets-manager.ts` (`c71bef10`), and neither is an ancestor of the other:

| Date | Commit | Event |
| --- | --- | --- |
| 2024-07-19 | `3664ed06` | #2660 authored. At that time the else-branch had a **single** `sync()` call; the PR adds `dot: true` to it. |
| 2024-08-09 | `793f3193` | #2687 merged (*"copy nested files and directories when no wildcard"*). It **replaces** that one call with two — outer `matchedPaths` + inner <code>sync(\`${matched}/**/*\`)</code>. Its base predates #2660, so neither call has `dot`. |
| 2024-10-21 | `486ac60e` | Merge commit *"chore: resolve conflicts"* finally lands #2660. The resolution puts `dot: true` back on the **outer** call, but the inner expansion call — which #2660's author never saw — is left without it. |

So this is not something #2660 overlooked; it is a merge-resolution gap against a call site introduced while #2660 was open. #2653 looks fixed because its reproduction used `config/envs/*`, which takes the working path.

## What is the new behavior?

`dot: true` is passed at the third call site as well, so `nest build` and `nest build --watchAssets` copy the same set of files — the parity #2660 established.

Tests were added at three levels:

| File | What it pins |
| --- | --- |
| `test/lib/compiler/assets-manager.spec.ts` | The expansion call is made with `dot: true`, both files are copied, and `exclude` is still forwarded alongside it |
| `test/lib/compiler/helpers/glob.spec.ts` | On the real filesystem: `<dir>/**/*` includes dotfiles with `dot`, omits them without — the helper contract the fix depends on |
| `test/e2e/build.command.e2e-spec.ts` | A real `nest build` on a scaffolded app copies `.development` under a bare directory asset, and the wildcard form still does too |

With the one-line fix reverted, the bare-directory e2e test fails while the wildcard one still passes — exactly the asymmetry the issue describes.

```
[ ] Yes
[x] No
```

Worth flagging even so: the fix copies **strictly more files** than before, so it is technically an observable change. Anyone who was relying on `nest build` silently skipping dotfiles under a directory asset (a stray `.DS_Store`, or a `.env` they assumed was not shipping) will see new files appear in `dist`. It restores documented-intent parity and matches what the same config already does in watch mode, so I have raised it as a `fix` — but if you would rather it rode a minor, that is your call.

## Other information

**Affected versions.** Present in `@nestjs/cli@11.0.24` (current `latest`) and on `master` at `12.0.0-alpha.5`, verified independently against `glob@13.0.6` and against the current internal `globSync` helper.

**Two unrelated things I ran into while writing the e2e tests**, neither addressed here:

1. Three pre-existing tests in `build.command.e2e-spec.ts` fail on `master` — *"default tsc compiler"*, *"custom tsconfig path using `--path`"*, and *"`--watch` mode initial compilation"*. They expect `dist/main.js`, but the scaffolded app emits to `dist/src/main.js`: the new-app template now ships root-level `vitest.config.ts` / `vitest.config.e2e.ts`, which widen the computed rootDir to the project root. I confirmed these fail without my changes applied. Either the template or those expectations needs updating — happy to open a separate issue.

2. The scaffolded app carries its own published `@nestjs/cli` in `node_modules`, and the CLI hands off to that local copy, so an e2e test can silently exercise the released build instead of the working tree. `library-assets.e2e-spec.ts` already calls `removeLocalCli` for this reason; the new block here does too.

Because of (1), the new e2e assertions *locate* the copied `prod.env` under `dist` and check that `.development` sits beside it, rather than hard-coding an output layout that is currently in question.

---

Thanks to Claude Opus 5

### DIFF
--- a/lib/compiler/assets-manager.ts
+++ b/lib/compiler/assets-manager.ts
@@ -237,6 +237,7 @@ export class AssetsManager {
                 if (statSync(matched).isDirectory()) {
                   return globSync(`${matched}/**/*`, {
                     ignore: item.exclude,
+                    dot: true,
                   }).filter((file) => statSync(file).isFile());
                 }
                 return matched;
--- a/test/e2e/build.command.e2e-spec.ts
+++ b/test/e2e/build.command.e2e-spec.ts
@@ -41,9 +41,7 @@ describe('Build Command (e2e)', () => {
     expect(fileExists(distDir)).toBe(true);
     expect(fileExists(path.join(distDir, 'main.js'))).toBe(true);
     expect(fileExists(path.join(distDir, 'app.module.js'))).toBe(true);
-    expect(fileExists(path.join(distDir, 'app.controller.js'))).toBe(
-      true,
-    );
+    expect(fileExists(path.join(distDir, 'app.controller.js'))).toBe(true);
     expect(fileExists(path.join(distDir, 'app.service.js'))).toBe(true);
   });
 
@@ -71,9 +69,7 @@ describe('Build Command (e2e)', () => {
       );
 
       // dist should be produced
-      expect(fileExists(path.join(appPath, 'dist', 'main.js'))).toBe(
-        true,
-      );
+      expect(fileExists(path.join(appPath, 'dist', 'main.js'))).toBe(true);
     } finally {
       proc.kill();
     }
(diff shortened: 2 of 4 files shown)


kamilmysliwiec:
lgtm

Issue #3532, opened 2026-08-28
Support Node's native subpath imports (package.json#imports, #specifier)


Roman991:
# COPILOT generated 🤖

### Feature / bug

Node's native [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) — the `imports` field in `package.json`, consumed via `#specifier` — don't survive `nest build` or `nest start` in an ESM (`"type": "module"`) project. TypeScript has no problem with them: under `moduleResolution: "nodenext"` it resolves and type-checks `#lib/hello.service.js` correctly. The gap is entirely in what the Nest CLI does (or rather doesn't do) to produce `dist/`.

Node resolves a `#specifier` against the **nearest `package.json`**, walking up from the file doing the import — not the project root. Since `nest build` emits into a separate `dist/` directory, that directory needs its own `package.json` (with the `imports` targets rewritten to be `dist`-relative) for `#specifier` imports to keep resolving once compiled. Right now there is no supported way to get that file there:

```
$ rm -rf dist && npx nest start
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '...\src\lib\hello.service.js' imported from ...\dist\app.module.js
```

(Node falls back to the project-root `package.json`, resolving the *source* `.ts` file's path as `.js`, since `dist/package.json` was never created.)

### What's blocking it

The obvious userland workaround is a `postbuild` npm script that copies + rewrites `package.json` into `dist/` — and that works fine when the build runs through `npm run build`. But `nest start` / `nest start --watch` invoke the compiler directly through `@nestjs/cli`'s own API, not `npm run build`, so `postbuild` never fires there.

The other candidate mechanism, `compilerOptions.assets`, turns out to be unusable for this specific case too: a project's own `package.json` always lives one directory above `sourceRoot`, and pointing an asset at it corrupts the whole build rather than just failing to copy:

```json
// nest-cli.json
{ "sourceRoot": "src", "compilerOptions": { "assets": [{ "include": "../package.json", "outDir": "dist" }] } }
```

```
$ rm -rf dist && npx nest build
error TS5033: Could not write file '.../dist/main.js': ENOENT: no such file or directory, ...
Found 6 error(s).

$ file dist
dist: JSON text data      # dist is now a *file* containing package.json's contents, not a directory
```

Traced to source (`@nestjs/cli` 12.0.0):

[...]

PR #3533, opened 2026-08-28
fix(assets): don't corrupt outdir for includes outside sourceroot

Roman991:
### What

Two related fixes in `AssetsManager`/`actionOnFile`:

1. When an `assets[].include` path resolves outside `sourceRoot` (e.g. the project's own `package.json`, one directory above `src/`), the sourceRoot-relative segment stripping can consume the filename itself, collapsing the destination to `outDir`. This silently overwrote `outDir` with the asset's contents before the compiler ran, corrupting the entire build. Now throws a clear error instead.
2. `assets[].flat` was declared in `configuration.d.ts` and threaded through `filesToCopy`/`collectLibraryAssets`, but never actually consulted when resolving the destination — setting it had no effect. It's now honored: `join(outDir, basename(path))`, bypassing the sourceRoot-relative stripping entirely. This is the escape hatch that makes (1) actionable instead of just diagnostic.

### Why

Copying a project's own `package.json` into `dist/` is needed to support Node's native [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) (`imports` field, `#specifier`) in an ESM (`"type": "module"`) project: Node resolves `#specifier` against the nearest `package.json` walking up from the *compiled* file doing the import, and `dist/` doesn't have one on its own. `package.json` always lives outside `sourceRoot`, which made this fail — see minimal repro: https://github.com/Roman991/nest-subpath-imports-repro

### Before / after (repro's `nest-cli.json`)

```json
{ "sourceRoot": "src", "compilerOptions": { "assets": [{ "include": "../package.json", "outDir": "dist" }] } }
```

```
# before
$ rm -rf dist && npx nest build
error TS5033: Could not write file '.../dist/main.js': ENOENT: no such file or directory, ...
$ file dist
dist: JSON text data          # outDir is now a *file*

# after
$ rm -rf dist && npx nest build
error  An error occurred during the assets copying process. Refusing to copy ".../package.json" to "dist": the "include" path is not deep enough under "sourceRoot" for any path segment to survive stripping, so the file would overwrite the "outDir" directory itself. Move the file inside "sourceRoot", or set "flat": true on this "assets" entry to copy it into "dist" by its basename instead.
```

With `"flat": true` added to the same entry:

```
$ rm -rf dist && npx nest build   # exit 0
$ find dist -type f
dist/app.module.js
dist/lib/hello.service.js
dist/main.js
dist/package.json
```

Verified end-to-end against the linked repro (with `rootDir: "."` added there so `dist/`he same `package.json` works unmodified for both the source tree and the compiled output — no custom `postbuild` script needed): `nest build` + `node dist/src/main.js`, `nest start`, and `nest start --watch` all correctly resolve the `#lib/*` subpath import.

### Backward compatibility

`copyPathResolve` itself is untouched — the existing unit test asserting it returns `outDir` when all segments are stripped (`copy-path-resolve.spec.ts`) still passes unchanged. The new behavior is a guard
at the call site in `actionOnFile`, which is where that return value was actually being

### Tests

Added to `test/lib/compiler/assets-manager.spec.ts`: throws instead of overwriting `outDir` when stripping consumes the whole path; does not throw when `flat` is set for the same case; `flat` copies to `outDir` by basename, bypassing `copyPathResolve` entirely; a `flat` destination is still confined to the project directory (`allowOutsidePaths` respected).

`npx vitest run` — 989 passing; the 3 pre-existing failures (`tsconfig-provider.spec.ts`s`) are present identically on `master` before this change and unrelated to`assets-manager`/`copy-path-resolve`.

### DIFF
--- a/lib/compiler/assets-manager.ts
+++ b/lib/compiler/assets-manager.ts
@@ -1,6 +1,6 @@
 import * as chokidar from 'chokidar';
 import { copyFileSync, mkdirSync, rmSync, statSync } from 'fs';
-import { dirname, join, sep } from 'path';
+import { basename, dirname, join, resolve, sep } from 'path';
 import {
   ActionOnFile,
   Asset,
@@ -343,11 +343,29 @@ export class AssetsManager {
     // Set path value to true for watching the first time
     this.watchAssetsKeyValue[assetCheckKey] = true;
 
-    let dest = copyPathResolve(
-      path,
-      item.outDir!,
-      sourceRoot.split(sep).length,
-    );
+    // "flat" bypasses the sourceRoot-relative stripping entirely and copies
+    // the file to "outDir" by its basename. This is the only way to target an
+    // "include" path that lives outside "sourceRoot" (e.g. the project's own
+    // package.json, needed one directory up for Node's subpath imports to
+    // resolve against the compiled output) without hitting the stripping
+    // logic below, which assumes the file is somewhere under "sourceRoot".
+    let dest = item.flat
+      ? join(item.outDir!, basename(path))
+      : copyPathResolve(path, item.outDir!, sourceRoot.split(sep).length);
+
+    // When "include" resolves outside "sourceRoot" (and "flat" isn't set),
+    // stripping sourceRoot's segment count from the file's path can consume
+    // the filename itself, collapsing the destination to "outDir" — silently
+    // overwriting the output directory with the asset's contents before the
+    // compiler even runs. Fail loudly instead.
+    if (!item.flat && resolve(dest) === resolve(item.outDir!)) {
+      throw new Error(
+        `Refusing to copy "${path}" to "${item.outDir}": the "include" path is not deep enough under ` +
+          `"sourceRoot" for any path segment to survive stripping, so the file would overwrite the ` +
+          `"outDir" directory itself. Move the file inside "sourceRoot", or set "flat": true on this ` +
+          `"assets" entry to copy it into "${item.outDir}" by its basename instead.`,
+      );
+    }
 
     // The destination is re-checked per file: the configured "outDir" is
     // validated up front, but a symlinked directory *inside* it would still
--- a/test/lib/compiler/assets-manager.spec.ts
+++ b/test/lib/compiler/assets-manager.spec.ts
@@ -727,6 +727,91 @@ describe('AssetsManager', () => {
     });
   });
 
+  describe('outDir collapse guard', () => {
+    it('throws instead of overwriting outDir when stripping consumes the whole path', () => {
+      // "../package.json" resolves outside "sourceRoot" (src/); stripping
+      // sourceRoot's segment count from it consumes the filename too, so
+      // copyPathResolve would otherwise return "dist" itself.
+      vi.mocked(globSync).mockReturnValue(['/cwd/package.json']);
+      vi.mocked(copyPathResolve).mockReturnValue('dist');
+      vi.mocked(getValueOrDefault)
+        .mockReturnValueOnce([{ include: '../package.json' }]) // assets
+        .mockReturnValueOnce([]) // includeLibraryAssets
+        .mockReturnValueOnce('src') // sourceRoot
+        .mockReturnValueOnce(false) // compilerOptions.watchAssets
+        .mockReturnValueOnce(undefined); // compilerOptions.allowOutsidePaths
+
+      expect(() =>
+        assetsManager.copyAssets({} as any, undefined, 'dist', false),
+      ).toThrow(/would overwrite the "outDir" directory itself/);
+
+      expect(copyFileSync).not.toHaveBeenCalled();
+
+      vi.mocked(copyPathResolve).mockReturnValue(
+        `${process.cwd()}/dist/file.txt`,
+      );
+    });
+
+    it('does not throw when flat is set, even for a path outside sourceRoot', () => {
+      vi.mocked(globSync).mockReturnValue(['/cwd/package.json']);
+      vi.mocked(copyPathResolve).mockReturnValue('dist');
+      vi.mocked(getValueOrDefault)
+        .mockReturnValueOnce([{ include: '../package.json', flat: true }]) // assets
+        .mockReturnValueOnce([]) // includeLibraryAssets
+        .mockReturnValueOnce('src') // sourceRoot
+        .mockReturnValueOnce(false) // compilerOptions.watchAssets
+        .mockReturnValueOnce(undefined); // compilerOptions.allowOutsidePaths
+
+      expect(() =>
+        assetsManager.copyAssets({} as any, undefined, 'dist', false),
+      ).not.toThrow();
+
+      expect(copyFileSync).toHaveBeenCalled();
+
+      vi.mocked(copyPathResolve).mockReturnValue(
+        `${process.cwd()}/dist/file.txt`,
+      );
+    });
+  });
+
+  describe('flat option', () => {
+    it('copies to outDir by basename, bypassing sourceRoot-relative stripping', () => {
+      vi.mocked(globSync).mockReturnValue(['/cwd/package.json']);
+      vi.mocked(getValueOrDefault)
+        .mockReturnValueOnce([{ include: '../package.json', flat: true }]) // assets
+        .mockReturnValueOnce([]) // includeLibraryAssets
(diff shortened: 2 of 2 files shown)


micalevisk (commented):
Thanks for the detailed writeup and the minimal repro — the `dist`-becomes-a-file behavior is real and nasty, and the root-cause trace in #3532 is accurate.

I checked the branch out and exercised the real `AssetsManager` against a real filesystem (not the spec's mocks), comparing this branch against its merge base. Full suite on the branch is green here: **86 files / 998 tests, 0 failures**, `tsc --noEmit` clean, Prettier clean. (The 3 pre-existing failures you mention don't reproduce on my machine — likely environment-specific, not a concern.)

Results, `sourceRoot: "src"`:

| scenario | before | this PR |
|---|---|---|
| `{include: "../package.json"}` | `dist` becomes a **file** | throws clear error ✅ |
| same, watch mode | `dist` becomes a **file** | error logged, build continues ✅ |
| same + `flat: true` | flat ignored → corruption | `dist/package.json` ✅ |
| `{include: "cfg/*.json", flat: true}` (inside sourceRoot) | `dist/cfg/x.json` | **`dist/x.json`** ⚠️ |
| `{include: "../other/**/*", flat: true}`, two `a.txt` | `dist/a.txt` = `A` | **`dist/a.txt` = `DEEP-A`** ⚠️ |
| `{include: "../other/**/*"}` (no flat) | `dist/a.txt`, `dist/deep/b.txt` | unchanged — still silently mis-mapped |
| `{include: "../package.json"}` + tsconfig `rootDir: "."` | **`dist/package.json`** ✅ | `dist/package.json` ✅ |

**The guard (change 1) looks right to me and I'd merge it.** The `flat` half (change 2) is where I'd push back.

---

### 1. `flat` isn't needed for the subpath-imports use case — `rootDir: "."` already works

That last table row is the key one. `build.action.ts` passes `tsRootDir = getEffectiveRootDir(...)` into `copyAssets`, and `assets-manager.ts` uses it as the strip root:

```ts
const stripRoot = rootDir ? join(rootDir) : sourceRoot;
```

So with tsconfig `rootDir: "."`, stripping happens relative to the **project root**, and `include: "../package.json"` maps to `dist/package.json` — on `master`, unpatched, with no `flat`. I confirmed this end-to-end.

Your own verification notes say the repro needed `rootDir: "."` added anyway ("so the same `package.json` works unmodified for both the source tree and the compiled output"). That change alone resolves #3532. The collapse only happens when `rootDir` is unset and inferred as `src` — in which case the emitted JS isn't `dist`-root-relative either, so `dist/package.json` wouldn't be where Node wants it regardless.

That means change 1 stands on its own: turn the silent corruption into an error whose message points at `rootDir` (or "move the file inside `sourceRoot`") rather than at `flat`.

### 2. Honoring `flat` is a silent breaking change for existing configs

`assets[].flat` has been accepted-but-ignored since it was introduced in e4aeecc (Mar 2020, v7) — six years across v7→v12. It's typed in `configuration.d.ts`, threaded through `filesToCopy`, and `assets-manager.ts` literally comments `// deprecated field` on it.

Any project with `"flat": true` sitting in `nest-cli.json` today gets a **different `dist/` layout after upgrading**, with no warning:

- `dist/cfg/x.json` → `dist/x.json` (row 4)
- basename collisions across directories now silently overwrite, last-writer-wins (row 5 — `dist/a.txt` flipped from `A` to `DEEP-A`)

Because the field has always been silently accepted, there's no way to distinguish "user meant this" from "user copy-pasted it years ago". The PR's Backward compatibility section only covers `copyPathResolve`'s unit test; this is the actual compat surface.

If maintainers do want `flat` alive, I'd want to see it land separately with: a breaking-change note, a warning on basename collision, removal of the `// deprecated field` comment (the code currently calls it deprecated while making it the sanctioned fix), and a docs update — `flat` isn't documented on docs.nestjs.com today, so nobody would discover the escape hatch anyway.

### 3. The guard is narrower than #3532's own suggestion

Row 6: `{include: "../other/**/*"}` still silently lands at `dist/a.txt` and `dist/deep/b.txt` — the outside-`sourceRoot` prefix is just dropped, and those destinations **collide with what `src/other/*` would emit**. Same class of bug, one segment shallower.

Suggestion #1 in the issue (`path.relative(sourceRootAbsolute, inPath)`, error when the result starts with `..`) catches the whole family; the `dest === outDir` check catches one member of it. Fine as deliberate minimal scope, but the PR title says "includes outside sourceroot" and it currently covers one case of that. Worth a follow-up at least.

Relatedly: the `depth(inPath) < up - 1` off-by-one in `copy-path-resolve.ts` is left in place. Correct call to guard at the call site instead, but that branch is now provably unreachable — worth deleting in a follow-up so it doesn't read as live protection.

### 4. Watch mode fails soft, non-watch fails hard

Rows 1 vs 2: in watch mode the throw is swallowed by `safeActionOnFile` and logged, so `nest build --watchAssets` prints the error and **exits 0**, while plain `nest build` fails. Not tearing down the watcher is right, but the severity asymmetry means CI on a watch-mode build won't catch the misconfiguration. Worth either a non-zero exit signal or an explicit note that it's warn-only there.

### 5. The new tests don't exercise the path arithmetic

`assets-manager.spec.ts` mocks `copyPathResolve`, and the "throws instead of overwriting outDir" test does:

```ts
vi.mocked(copyPathResolve).mockReturnValue('dist');
```

That asserts the guard fires *given a fabricated collapse* — it never asserts that the real arithmetic collapses for `../package.json`. If `copyPathResolve` changed, the test stays green while the guard goes dead. I realize that's the file's house style, but a real-fs case (or an e2e entry) would pin the actual behavior.

Minor hygiene: both new guard tests restore the `copyPathResolve` mock inline at the end of the test body. `beforeEach` uses `clearAllMocks()`, which does not reset implementations — so if the `toThrow` assertion ever fails, the restore is skipped and `'dist'` leaks into every subsequent test in the file. Move it to an `afterEach`.

### Nit

`import { resolve }` from `path` now shadows the `new Promise<void>((resolve) => …)` parameter in `copyAssets`. Different method, no functional bug, but worth avoiding in a file this path-heavy.

---

### Suggested path forward

1. **Merge the guard**, with the error message pointing at `rootDir` (the fix that actually works today) instead of `flat`.
2. **Split the `flat` change out** into its own PR/discussion — it's a six-year-dormant deprecated option, resurrecting it changes existing builds silently, and per §1 it isn't required for #3532.
3. Optional follow-up: the `path.relative` version of the check, to cover §3, plus deleting the now-unreachable `depth()` guard.

kamilmysliwiec:
Thanks for this, and for the minimal repro — the `dist`-becomes-a-file behavior is a genuinely nasty failure mode, and the root-cause trace in #3532 is accurate.

I've taken the guard with a few changes; it's landing on `master`:

**The guard is in.** Refusing the copy is the right call — silently replacing the output directory with an asset before the compiler runs is about as bad as a build bug gets.

**The error message now points at `rootDir` rather than `flat`.** `copyAssets` already strips relative to the effective tsconfig `rootDir` when one is set, so `rootDir: "."` maps `include: "../package.json"` to `dist/package.json` on `master` today, unpatched. Your own verification notes mention adding `rootDir: "."` to the repro anyway — that alone resolves #3532, which is why the guard can stand on its own.

**I left the `flat` half out.** `assets[].flat` has been accepted-but-ignored since v7 — it's typed, threaded through `filesToCopy`, and the code literally comments `// deprecated field` on it. Honoring it now would silently change `dist/` layout for any project that has it sitting in `nest-cli.json`, including basename collisions across directories resolving last-writer-wins, and it flows into library assets too. If we want `flat` alive it should land on its own with a breaking-change note, a collision warning, and docs — happy to discuss in a separate issue.

**Two other notes:**

- I kept `copyPathResolve`'s `depth(inPath) < up - 1` branch. It isn't dead code — it covers include paths with *fewer* segments than the strip root, which the new `dest === outDir` check doesn't reach.
- The specs resolve `lib/**/*.js` to the compiled build output rather than the `.ts` sources, so the guard silently did nothing under test until I ran `npm run build`. That's also what your 3 "pre-existing failures" were — after a rebuild the suite is fully green. The landed test uses the real `copyPathResolve` via `importActual` instead of mocking a fabricated collapse, so it pins the actual path arithmetic.

Credited you in the commit trailer. Thanks again for chasing this down.

PR #3535, opened 2026-08-29
fix(compiler): support ESM and CJS rspack config

id6tm-mteterel:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When building with Rspack and providing a custom configPath, the CLI imported it by calling `require` leading up to a silent failure when the file is an ES Module.

## What is the new behavior?

The call moves from `require` to use `async import` instead, which allows both ESM and CJS config files to be loaded. Tests have been added with fixtures for both scenarios.

```
[x] Yes
[ ] No
```

Consumers in ESM mode (`type: module`) using a custom CJS rspack config with a `.js` extension will have to either change it to `.cts` or convert it to ESM syntax.

### DIFF
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -34,6 +34,11 @@ import type webpack from 'webpack';
 
 const require = createRequire(import.meta.url);
 
+type RspackConfigFactory = Record<string, any> | ((
+  config: Record<string, any>,
+  rspackRef: any,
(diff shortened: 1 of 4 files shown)


id6tm-mteterel:
I'll close this as #3553 has broader support

PR #3542, opened 2026-09-02
fix(compiler): resolve tsconfig absolute path

id6tm-mteterel:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When using the Nest CLI Rspack builder in a project with an inherited TypeScript configuration, the tsconfig path is passed to Rspack as a relative path. Rspack may fail to resolve extended configurations outside the application directory, resulting in errors such as:

```
Tsconfig not found tsconfig.root.json
```

The same project builds successfully with the TypeScript builder.

## What is the new behavior?

The Nest CLI now resolves the TypeScript configuration path to an absolute path before passing it to Rspack and TsconfigPathsPlugin. This allows inherited tsconfig files to resolve correctly in monorepos and nested project layouts.

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/compiler/rspack-compiler.ts
+++ b/lib/compiler/rspack-compiler.ts
@@ -1,6 +1,6 @@
 import { existsSync } from 'fs';
 import { createRequire } from 'module';
-import { join } from 'path';
+import { resolve } from 'path';
 import type * as ts from 'typescript';
 import { Configuration } from '../configuration/index.js';
 import { ERROR_PREFIX, INFO_PREFIX } from '../ui/index.js';
@@ -44,7 +44,7 @@ export class RspackCompiler extends BaseCompiler<RspackCompilerExtras> {
     onSuccess?: () => void,
   ) {
     const cwd = process.cwd();
-    const configPath = join(cwd, tsConfigPath!);
+    const configPath = resolve(cwd, tsConfigPath!);
     if (!existsSync(configPath)) {
       throw new Error(
         `Could not find TypeScript configuration file "${tsConfigPath!}".`,
@@ -72,7 +72,7 @@ export class RspackCompiler extends BaseCompiler<RspackCompilerExtras> {
       entryFileRoot,
       entryFile,
       extras.debug ?? false,
-      tsConfigPath,
+      configPath,
       plugins,
       isEsmProject(),
       extras.tsOptions,
--- a/test/lib/compiler/rspack/rspack-compiler.spec.ts
+++ b/test/lib/compiler/rspack/rspack-compiler.spec.ts
@@ -11,6 +11,7 @@ import { PluginsLoader } from '../../../../lib/compiler/plugins/plugins-loader.j
 import { RspackCompiler } from '../../../../lib/compiler/rspack-compiler.js';
 
 import { existsSync } from 'fs';
+import { resolve } from 'path';
 import { rspackDefaultsFactory } from '../../../../lib/compiler/defaults/rspack-defaults.js';
 import { getValueOrDefault } from '../../../../lib/compiler/helpers/get-value-or-default.js';
 import * as esmProjectUtil from '../../../../lib/utils/is-esm-project.js';
@@ -71,6 +72,7 @@ vi.mock('../../../../lib/compiler/helpers/get-value-or-default.js', () => ({
 vi.mock('../../../../lib/utils/is-esm-project.js');
 
 describe('Rspack Compiler', () => {
+  const absoluteTsconfigPath = resolve(process.cwd(), 'tsconfig.json');
   let compiler: RspackCompiler;
 
   // Access mock functions from the hoisted rspack mock
@@ -154,7 +156,7 @@ describe('Rspack Compiler', () => {
       ).toThrow('Could not find TypeScript configuration file');
     });
 
-    it('should call rspackDefaultsFactory with correct arguments', () => {
+    it('should pass an absolute tsconfig path to Rspack', () => {
       vi.mocked(existsSync).mockReturnValue(true);
       vi.mocked(getValueOrDefault)
         .mockReturnValueOnce('main') // entryFile
@@ -175,7 +177,7 @@ describe('Rspack Compiler', () => {
         '',
         'main',
         false,
-        'tsconfig.json',
+        absoluteTsconfigPath,
         expect.objectContaining({
           beforeHooks: [],
           afterHooks: [],
@@ -205,7 +207,7 @@ describe('Rspack Compiler', () => {
         '',
         'main',
         false,
-        'tsconfig.json',
+        absoluteTsconfigPath,
         expect.anything(),
         true,
         undefined,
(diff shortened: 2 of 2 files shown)


KayanoLiam:
This also fixes nestjs/nest#17711 in a minimal monorepo reproduction. I compared CLI master `cc0b976c` with this PR's current head, `12adaa59989da8c4202476389c2da596235741c6`, using Rspack 2.2.3 and Node.js 26.7.0.

The setup has `apps/admin` and `apps/worker` importing a service through `@app/core`, mapped in the root tsconfig to `libs/core/src/index.ts`. After the initial watch build, I changed a comment in the service and then changed its returned string, checking the emitted JavaScript after each successful rebuild.

- On master, both application bundles retain the original comment and returned string. With `--all --watch`, the library's own bundle updates while its consuming applications stay stale.
- With this PR, both edits reach every bundle without restarting the CLI. I verified `build admin --watch`, `build --all --watch`, and `build --all --watch` with an ESM project.

Validation also passed: `npm run build` and the three existing Rspack test files (56 tests), via `npm test -- test/lib/compiler/rspack test/lib/compiler/defaults/rspack-defaults.spec.ts`.

Passing the absolute tsconfig path therefore addresses the reported watch regression as well as the inherited-config resolution issue described above.

lib/compiler/rspack-compiler.ts
```diff
@@ -44,7 +44,7 @@ export class RspackCompiler extends BaseCompiler<RspackCompilerExtras> {
     onSuccess?: () => void,
   ) {
     const cwd = process.cwd();
-    const configPath = join(cwd, tsConfigPath!);
+    const configPath = resolve(cwd, tsConfigPath!);
```
micalevisk, line 47: `resolve` is the right call here: besides normalising, it keeps an already-absolute `tsConfigPath` intact (with `join` you'd get `/cwd//abs/tsconfig.json`). Just noting for the record that this branch is currently unreachable with an absolute value, because `TsConfigProvider.getByConfigFilename` (called earlier in `build.action.ts`) still uses `join(process.cwd(), …)` and throws first. Worth aligning in a follow-up so `--path /abs/tsconfig.json` works for all builders.

lib/compiler/rspack-compiler.ts
```diff
@@ -72,7 +72,7 @@ export class RspackCompiler extends BaseCompiler<RspackCompilerExtras> {
       entryFileRoot,
       entryFile,
       extras.debug ?? false,
-      tsConfigPath,
+      configPath,
```
micalevisk, line 75: Verified this is the line that fixes both symptoms. `configPath` fans out to `resolve.tsConfig`, `TsconfigPathsPlugin({ configFile })` and `ForkTsCheckerWebpackPlugin({ typescript: { configFile } })`; all three accept absolute paths. With the relative value, Rspack's native resolver was recording alias-resolved modules as relative resources/`fileDependencies` (`libs/core/src/index.ts`), which never match the absolute paths emitted by the watcher, hence the stale bundles in nestjs/nest#17711 and the broken `extends` base dir for `../tsconfig.root.json`.

test/lib/compiler/rspack/rspack-compiler.spec.ts
```diff
@@ -154,7 +156,7 @@ describe('Rspack Compiler', () => {
       ).toThrow('Could not find TypeScript configuration file');
     });
 
-    it('should call rspackDefaultsFactory with correct arguments', () => {
+    it('should pass an absolute tsconfig path to Rspack', () => {
```
micalevisk, line 159: nit: this test still asserts every argument of `rspackDefaultsFactory`, so the old name ("should call rspackDefaultsFactory with correct arguments") described it better. I'd keep that name and add a small dedicated case that documents *why* `resolve` was chosen over `join`, e.g. calling `run()` with an absolute `tsConfigPath` and asserting it is forwarded unchanged. Not blocking.

micalevisk (approved):
Reviewed against head `12adaa5` with a local monorepo reproduction (Rspack 2.2.1, TS 6, Node 24). **Approving.** The change is minimal, correct, and matches Rspack's documented usage of `resolve.tsConfig` (every example in the docs builds it with `path.resolve(...)`).

### What I verified

| Scenario | master (`4498d54`) | this PR |
|---|---|---|
| `tsconfig.json` with `"extends": "../tsconfig.root.json"` (outside cwd), `nest build` | ❌ `Tsconfig not found tsconfig.root.json` | ✅ compiles |
| Monorepo, `nest build admin --watch`, edit `libs/core` imported via `@app/core` path alias | ❌ `dist/apps/admin/main.js` stays stale after "compiled successfully" | ✅ bundle updated |
| Same, `nest build --all --watch` | ❌ `dist/libs/core` updates, `dist/apps/admin` stays stale (exactly nestjs/nest#17711) | ✅ both updated |
| `nest build -p ./tsconfig.json` | ✅ | ✅ |

So the report from @KayanoLiam is accurate: this PR also fixes nestjs/nest#17711, and the root cause is the same. With a relative `resolve.tsConfig`, Rspack's native resolver registers alias-resolved modules with a **relative** resource and a relative `fileDependencies` entry (`libs/core/src/index.ts`), while the watcher reports absolute paths, so the module is never invalidated. With the absolute path both are absolute. Dumped from `stats.compilation` via a throwaway `rspack.config.js`:

```
master: MOD libs/core/src/index.ts            FILEDEP libs/core/src/index.ts
PR:     MOD /abs/.../fxB/libs/core/src/index.ts  FILEDEP /abs/.../fxB/libs/core/src/index.ts
```

Non-regression: full suite green locally (86 files / 994 tests), `tsc --noEmit` clean, CI green. The only observable change for users with a custom `rspack.config.js` factory is that `options.resolve.tsConfig` (and the `configFile` handed to `TsconfigPathsPlugin` / `ForkTsCheckerWebpackPlugin`) is now absolute, which all three consumers accept. Not a breaking change. Prettier flags line 27 of `rspack-compiler.ts`, but that line is untouched and master fails the same check, so it's unrelated drift.

Prior art: no earlier PR or issue in this repo attempted this, and there is no maintainer "won't fix" on the topic. The Rspack builder has only existed since v12 (`84e103e`), and the relative path was introduced there without discussion.

### Follow-ups (not blocking, happy to open issues)

1. `TsConfigProvider.getByConfigFilename` still does `join(process.cwd(), configFilename)`, so an **absolute** `--path` / `tsConfigPath` fails there before ever reaching this compiler (`Could not find TypeScript configuration file "/abs/..."`). The `join → resolve` switch here is therefore only latent today; aligning the provider (and the deprecated `webpack-compiler.ts:64`) would make absolute paths work end to end for every builder.
2. `rspack-defaults.spec.ts` never asserts that `tsConfigFile` reaches `resolve.tsConfig` / `TsconfigPathsPlugin({ configFile })`. A one-line assertion there would guard the seam this PR depends on.

kamilmysliwiec:
lgtm

PR #3404, opened 2026-04-17
fix(plugin-metadata): emit .js extensions for dynamic imports under nodenext (#3364)

maruthang:
## PR Checklist
- [x] The commit message follows our guidelines: https://github.com/nestjs/nest-cli/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
Issue Number: #3364

`PluginMetadataGenerator` printed visitor-supplied dynamic `import("./hello.dto")` `ts.CallExpression`s and the `typeImports` strings verbatim. Under `moduleResolution: 'node16' | 'nodenext'`, ESM specifiers must include explicit file extensions — so the generated `metadata.ts` failed type-check and threw `ERR_MODULE_NOT_FOUND` at runtime.

Two prior attempts (#3378, #3365) tried to suppress the error with `@ts-nocheck`; both were closed because the maintainer wanted the underlying generation fixed instead.

## What is the new behavior?
The generator now reads `programRef.getCompilerOptions().moduleResolution`. When it's `Node16` or `NodeNext`, it walks the metadata tree:
- For each `import()` `CallExpression` whose specifier is a relative path without an extension, rewrite the specifier to append `.js`.
- For the `typeImports` map (raw strings of the shape `await import(...)`), regex-rewrite the inner `import(\"...\")` specifier the same way.

Bare specifiers, absolute paths, and already-extended specifiers (`.js`, `.mjs`, `.cjs`, `.json`, `.ts`, `.tsx`, etc.) are skipped, so classic `node10` and `bundler` resolution are unaffected.

Five small pure helpers are exported alongside the rewrite logic, covered by 18 new unit tests in `test/lib/compiler/plugins/plugin-metadata-generator.spec.ts`. Full local suite: 21 suites, 166 passed, 3 pre-existing skips, 0 failures.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

The rewrite only activates when `moduleResolution` is `Node16` or `NodeNext`. Default `node` resolution behavior is identical to before.

## Other information
[...]

Issue #2858, opened 2025-01-06
tsconfig-paths does not work with ESM; possible switch to tsc-alias


dyedwiper:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

tsconfig-paths is used by the nest-cli to rewrite paths according to path aliases specified in tsconfig.json.

tsconfig-paths does not work with ES Modules as reported here e.g.:
- https://stackoverflow.com/questions/71571684/ts-node-with-tsconfig-paths-wont-work-when-using-esm
- https://github.com/dividab/tsconfig-paths/issues/243
- https://github.com/TypeStrong/ts-node/issues/1375

Thus `nest start` throws an ERR_MODULE_NOT_FOUND when compiling to ES Modules and using path aliases. Here is a minimal example: https://stackblitz.com/edit/nestjs-typescript-starter-m27m4jmo

It is related to this issue: https://github.com/nestjs/nest-cli/issues/1437. Though AFAIU the problem is not caused by the file extension. Thus I decided to open a new issue. Hope that's fine.

### Describe the solution you'd like

A solution could be to use [tsc-alias](https://github.com/justkey007/tsc-alias) as a post-compilation step, instead of using tsconfig-paths in pre-compilation.

A possible benefit: Rewriting the paths with tsconfig-paths in pre-compilation involves some [custom code](https://github.com/nestjs/nest-cli/blob/master/lib/compiler/hooks/tsconfig-paths.hook.ts#L75) which you could probably lose with tsc-alias.

I haven't fully tested it, but when I change the build script to `"nest build && tsc-alias -p tsconfig.build.json"`, the import is correct in the compiled `dist/app.module.js` after running `npm run build`. See here: https://stackblitz.com/edit/nestjs-typescript-starter-rcmvukfp

### What is the motivation / use case for changing the behavior?

We would like to use path aliases in a project and this hinders it at the moment.

SrHadeSito:
+1

ngothanhluc:
+1

Issue #1437, opened 2021-12-02
tsconfig-paths rewrite to be broken when import with file extension (.js)


msimon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

With a project organised that way:
```
src:
  - main.ts
  - app.module.ts
  ....
tsconfig.json
...
```

Defining an import in `main.ts` the following way:
```
import { AppModule } from 'src/app.module.js';
```
result in it being compiled to:
```
const app_module_1 = require("src/app.module.js");
```
ignoring the non-relative path to relative path rewrite that should be done by `tsconfig-paths`.

Removing the `.js` from the import, result in the correct compilation:
```
import { AppModule } from 'src/app.module';
```
```
const app_module_1 = require("./app.module");
```

I took a quick look into `nest-cli/lib/compiler/hooks/tsconfig-paths.hook.ts` to see if I could spot the issues.
My guess is that the match on l.56 does not return anything if the extension is set.

I'll try to ding more into it soon, but I figure I would ask in case there is an obvious fix for this.

### Minimum reproduction code

.

### Steps to reproduce

After adding/removing the .js
```
nest build --tsc
cat dist/main.js
```

### Expected behavior

Defining an import in `main.ts` the following way:
```
import { AppModule } from 'src/app.module.js';
```
Should compile to:
```
const app_module_1 = require("./app.module.js");
```

### Package version

8.1.5

### NestJS version

8.2.3

### Node.js version

v14.17.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

jmcdo29:
Can you please provide a minimum reproduction repository?

kamilmysliwiec:
BTW using absolute paths (using "src" in your import paths) is considered as a bad practice

msimon:
@jmcdo29 https://github.com/HomeskilletHealthInc/nestJs_import_rewrite_bug

@kamilmysliwiec Using absolute path is, but I don't believe that applies to non-relative path. The code will always compile to the same relative pass, independently of the service/computer is being compiled on. 
The gain of using non-relative import is so much higher imo. It's even supported out of the box by NestJs.

Let me know if I misses something, but I haven't seen any large projects using fully relative path, nor a good argument to use them.

kamilmysliwiec:
Yeah just to clarify: it applies only to paths that contain `src` directory (so using `rootDir` + absolute paths in your tsconfig)

[...]

PR #3545, opened 2026-09-02
fix(compiler): emit loadable path alias specifiers for esm

johnnyhuirilef:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A (see nestjs/nest#17650)

Under `node16`/`nodenext`, the tsconfig paths hook emits module specifiers Node
cannot load. Two forms:

**An alias that resolves to a directory** is emitted as a directory import:

```js
import { MyLibraryModule } from "../../../libs/my-library/src";
// ERR_UNSUPPORTED_DIR_IMPORT
```

**An alias written with the extension the file carries after emit** is not
rewritten at all. The matcher looks for `./service.js` on disk, finds only
`./service.ts`, and returns undefined, so the alias is emitted untouched:

```js
import { Service } from "@app/my-library/service.js";
// ERR_MODULE_NOT_FOUND
```

`tsconfig-paths` drops the extension it resolved and collapses an index hit to
its directory, so the path it returns is never a usable ESM specifier.

A generated monorepo reaches the first form without writing anything: `nest g
library` writes an alias that points at `libs/<name>/src`.

## What is the new behavior?

A directory alias points at its index file, and a failed match is retried
without the output extension.

```
                                 before                     after
directory alias                  "./lib/src"                "./lib/src/index.js"
alias with emitted extension     left as the alias          "./lib/src/service.js"
commonjs                         "./lib/src"                "./lib/src"
```

Both are gated on `requiresExplicitImportExtensions()`, the helper the plugin
metadata generator already uses for the same question (#3404). CommonJS output
is byte-identical.

```
[ ] Yes
[x] No
```

## Other information

**Scope.** The `tsc` builder only. rspack resolves aliases itself and never
calls this hook.

**Tests.** Four in `tsconfig-paths.hook.spec.ts` with a new fixture. The three
ESM ones fail without the change; the CommonJS one passes either way and covers
that path. `.mts` and `.cts` targets keep their own output extension.

**Context.** #1437 reported the second form in 2021, and #2858 asked for the
same under ESM. Both were closed alongside the v12 release, which brought this
hook to ESM but left the resolution as it was. Both forms still reproduce on
master.

**One thing worth a look:** `requiresExplicitImportExtensions` lives in
`plugins/plugin-metadata-generator.ts`, so importing it here is the only
`hooks` → `plugins` reference in the tree. Happy to move it to a shared module
if you prefer that.

**What this does not fix.** The `TS2307` a type-checker reports for a directory
alias under `nodenext`. That comes from the generated `paths` value, not from
this hook.


lib/compiler/hooks/tsconfig-paths.hook.ts
```diff
@@ -87,17 +90,67 @@ function getModuleSpecifierText(
...
+ * Gives the specifier the extension its emitted file will carry, which ESM
+ * requires and the resolved source path does not always have.
+ */
+function withOutputExtension(specifier: string): string {
+  for (const [source, output] of Object.entries(SOURCE_TO_OUTPUT_EXTENSION)) {
+    if (specifier.endsWith(source)) {
+      return specifier.slice(0, -source.length) + output;
+    }
+  }
+  return posix.extname(specifier) ? specifier : `${specifier}.js`;
```
micalevisk: **Blocking.** `posix.extname()` treats `.service` / `.module` / `.controller` as an extension. `tsconfig-paths` strips only the last extension of the file it found, so `users.service.ts` arrives here as `./lib/users.service`, and this line returns it unchanged: `import ... from "./lib/users.service"` → `ERR_MODULE_NOT_FOUND`. Reproduced with `paths: { '~lib/*': ['./src/lib/*'] }` and `import { UsersService } from '~lib/users.service.js'`. See the review body for a verified replacement and please add a dotted-name fixture.

lib/compiler/hooks/tsconfig-paths.hook.ts
```diff
@@ -87,17 +90,67 @@ function getModuleSpecifierText(
   return importPathWithQuotes.substring(1, importPathWithQuotes.length - 1);
 }
 
+const MATCHER_EXTENSIONS = ['.ts', '.tsx', '.js', '.jsx'];
```
micalevisk: `~lib/mod.mjs` (with a `mod.mts` source) is still emitted untouched: the retry strips it to `~lib/mod` and this list has no `.mts`/`.cts` to probe. The `.mts → .mjs` mapping below only fires when the alias target names the file explicitly, which is what the new test does. Adding the extensions here alone isn't enough either, since the matcher strips `mod.mts` to `mod` and `withOutputExtension` would then append `.js`. The `toEmittedPath` shape in the review body handles both.

lib/compiler/hooks/tsconfig-paths.hook.ts
```diff
@@ -1,7 +1,9 @@
+import { existsSync } from 'fs';
 import { createRequire } from 'module';
 import { dirname, posix } from 'path';
 import * as tsPaths from 'tsconfig-paths';
 import * as ts from 'typescript';
+import { requiresExplicitImportExtensions } from '../plugins/plugin-metadata-generator.js';
```
micalevisk: Let's take you up on the offer: please move `requiresExplicitImportExtensions` to `lib/compiler/helpers/requires-explicit-import-extensions.ts` (matching the one-helper-per-file convention there) and import it from that path in both places, so `hooks/` doesn't depend on `plugins/`.

test/lib/compiler/hooks/tsconfig-paths.hook.spec.ts
```diff
@@ -280,4 +280,51 @@ describe('tsconfig paths hooks', () => {
       }
     });
   }, 15000);
+
+  describe('esm output', () => {
```
micalevisk: Nit: there's already a `describe('ESM output (module: ESNext)')` above; `'ESM output (module: NodeNext)'` keeps them symmetric, and the sibling blocks pass `15000` as timeout.

More important: under `module: NodeNext` TS picks CJS vs ESM from the nearest `package.json`, and the fixture has none, so these pass only because the repo root has `"type": "module"`. Please add `fixtures/esm-imports/package.json` with `{ "type": "module" }`. A `.d.ts` assertion (`declaration: true`) would also be worth it since the hook runs in `afterDeclarations` too.

micalevisk (changes requested):
Thanks for the PR, and for the thorough description. The direction is right: it matches what we settled on in #3404 (emit correct ESM specifiers, gated on `moduleResolution`), and it is the nest-cli half of the plan discussed in nestjs/nest#17706 and nestjs/schematics#2444. I ran the branch locally (spec, typecheck, lint, prettier all green) and also ran the hook against a few input shapes the new tests don't cover. One of them is a blocker.

### Blocking

**1. Dotted file names lose their `.js`** (`withOutputExtension`, L131)

`withOutputExtension` uses `posix.extname()` to decide whether the specifier already carries an extension. But `tsconfig-paths` returns an extension-probed hit with only the *last* extension stripped (`removeExtension` is `substring(0, lastIndexOf('.'))`), so `users.service.ts` comes back as `.../users.service`, `extname` reports `.service`, and nothing is appended:

```ts
// src/main.ts — paths: { '~lib/*': ['./src/lib/*'] }, module + moduleResolution: nodenext
import { UsersService } from '~lib/users.service.js';

// dist/main.js on this branch
import { UsersService } from "./lib/users.service";   // ERR_MODULE_NOT_FOUND
```

`*.service.ts`, `*.module.ts`, `*.controller.ts` are the mainstream Nest layout, and `@app/<lib>/*` is exactly the alias nestjs/schematics#2444 keeps for ESM, so this is the common path rather than an edge case. The new `service.ts` fixture doesn't catch it because the name has no dot. Please add a `users.service.ts` (or similar) to the fixture and assert `from "./lib/users.service.js"`.

### Should be addressed in this PR

**2. `.mjs` / `.cjs` aliases never resolve** (`MATCHER_EXTENSIONS`, L93)

`MATCHER_EXTENSIONS` has no `.mts`/`.cts`, so the retry in `matchPath` strips `~lib/mod.mjs` to `~lib/mod`, the matcher probes `mod.ts|tsx|js|jsx`, finds nothing, and the alias is emitted untouched. The `.mts → .mjs` entry in `SOURCE_TO_OUTPUT_EXTENSION` only fires when the alias *target itself* names the file (`'~mod': ['./src/mod.mts']`), which is what the new test does, so the mapping looks covered while the `~lib/*` form isn't. This is pre-existing on master, but this PR is the one introducing `.mts`/`.cts` awareness, so it should be consistent.

Note that just adding `.mts`/`.cts` to `MATCHER_EXTENSIONS` is not enough: the matcher would then strip `mod.mts` to `mod` and `withOutputExtension` would append `.js`.

**Suggested shape for 1 + 2.** The root cause of both is guessing the output extension from a path whose extension was thrown away. Since `resolveIndexFile` already probes the disk, one function can do both jobs: find which source file the matcher actually hit, then map its real extension. I verified this locally on this branch: all 13 tests in the spec pass, and it emits the right specifier for `users.service.js`, `mod.mjs`, a directory alias, a dotted directory (`my.lib`), and a `package.json#main` that points at a `.ts` file.

```ts
const SOURCE_TO_OUTPUT_EXTENSION: Record<string, string> = {
  '.mts': '.mjs',
  '.cts': '.cjs',
  '.ts': '.js',
  '.tsx': '.js',
  '.js': '.js',
  '.jsx': '.js',
};
const MATCHER_EXTENSIONS = Object.keys(SOURCE_TO_OUTPUT_EXTENSION);

/**
 * The matcher strips the extension of the file it found (and collapses an
 * index hit to its directory). Probe the disk once more to learn which
 * source file it was, and return it with the extension it will carry after emit.
 */
function toEmittedPath(resolved: string): string {
  for (const candidate of [resolved, posix.join(resolved, 'index')]) {
    for (const [source, output] of Object.entries(SOURCE_TO_OUTPUT_EXTENSION)) {
      if (candidate.endsWith(source) && existsSync(candidate)) {
        return candidate.slice(0, -source.length) + output;
      }
      if (existsSync(candidate + source)) {
        return candidate + output;
      }
    }
  }
  return resolved;
}
```

and in `getNotAliasedPath`:

```ts
  if (esm) {
    result = toEmittedPath(result);
  }
  const resolvedPath = posix.relative(dirname(sf.fileName), result) || './';
  return resolvedPath[0] === '.' ? resolvedPath : './' + resolvedPath;
```

`resolveIndexFile` and `withOutputExtension` go away. Feel free to shape it differently; the important part is that the extension comes from the file that exists on disk, not from the stripped string.

**3. Move `requiresExplicitImportExtensions` out of `plugins/`** (L6)

Yes please, in this PR. `lib/compiler/helpers/` is where it belongs (one kebab-case file per helper, e.g. `requires-explicit-import-extensions.ts`), with `plugin-metadata-generator.ts` importing it from there. That removes the `hooks → plugins` edge.

**4. Tests**

- `describe('esm output')` sits next to `describe('ESM output (module: ESNext)')`. Name it `'ESM output (module: NodeNext)'` for symmetry, and keep the `15000` timeout the sibling blocks have.
- Under `module: NodeNext`, TS decides CJS vs ESM from the nearest `package.json`. The fixture has none, so these tests pass only because the repo root `package.json` has `"type": "module"`. Add `fixtures/esm-imports/package.json` with `{ "type": "module" }` so the fixture is self-describing.
- The hook also runs in `afterDeclarations` (`compiler.ts`, `watch-compiler.ts`), so `.d.ts` output now gets `./lib/index.js` too, which is what nodenext wants. One assertion on `dist/main.d.ts` with `declaration: true` would lock that in.

### Follow-ups (not for this PR, noted so they don't get lost)

- `requiresExplicitImportExtensions` only looks at `moduleResolution`. A tsconfig with just `"module": "nodenext"` (the form the TS docs recommend) parses with `moduleResolution: undefined`, so the gate is off and none of this applies. Our scaffold sets both, so not urgent, but the helper should probably also accept `module: Node16 | NodeNext`. Pre-existing from #3404.
- A `package.json#main` that points at a `.ts` file is emitted with the `.ts` extension in CommonJS output (`require("../libs/pkg/src/index.ts")`) on master. This PR fixes that for ESM only.

### Small accuracy note on the description

"CommonJS output is byte-identical" holds when `moduleResolution` is not node16/nodenext. A `node16` project without `"type": "module"` emits CJS and now gets `require("./lib/index.js")` instead of `require("./lib")`. Still loadable, so fine, just not byte-identical.

### Verified

- `vitest run test/lib/compiler/hooks/tsconfig-paths.hook.spec.ts`: 13/13 pass. `tsc --noEmit`, `oxlint`, `prettier --check`: clean.
- Branch is 6 renovate commits behind master, mergeable, no conflicts.
- Prior art: #1437 (2021) and #2858 (2025) reported these same forms and were closed in the v12 sweep without a fix and without a "won't support". #3404 set the gating approach this PR reuses. nestjs/schematics#2444 is the schematics half and depends on this landing in the same release.

---
_Reviewed with help from Claude Code (Fable 5.1). Every claim above was reproduced locally on this branch._

johnnyhuirilef:
Thanks — the dotted-name case was the common one and my fixture walked right
past it.

I went with your suggestion and probed the disk. One function now finds which
source file the matcher landed on and gives it the extension it will carry after
emit, so the dotted case and the wildcard `.mjs` aliases fall out of the same
change.

`requiresExplicitImportExtensions` moved to `lib/compiler/helpers/`. The fixture
has its own `package.json`; I checked it by deleting `type` from the repo root
and the tests still pass. Added the `.d.ts` assertion, the 15000 timeout, and
the symmetric name.

Two more things surfaced while doing this:

**`.tsx` does not always emit `.js`.** It follows `jsx`:

```
jsx=preserve   -> cmp.jsx
jsx=react-jsx  -> cmp.js
jsx=react      -> cmp.js
```

The map had it fixed at `.js`, so under `preserve` the specifier pointed at a
file that does not exist — the same failure this PR exists to fix. The map is
derived from `compilerOptions.jsx` now.

**Four map entries had no test behind them.** I mutated each one against the
1003 and they survived: `.cts`, `.tsx`, `.jsx`, `.js`. They have tests now, and
no mutation gets through.

On `existsSync`: I measured 25 calls for 6 aliases, up to 9 on one. There is a
`Map` per transform now — resolution does not change within a build.

1003 tests.

Issue #3547, opened 2026-09-04
`nest generate --help` is unreadable on small terminals width


biisal:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`nest g --help` always renders the schematics list as a table (`name | alias | description`) even on terminals with smaller width/columns. This makes the output messy/overflows.

<img width="649" height="915" alt="Image" src="https://github.com/user-attachments/assets/c653ada1-08de-4c5a-915f-1b20ca841701" />

### Minimum reproduction code

https://github.com/user-attachments/assets/c653ada1-08de-4c5a-915f-1b20ca841701

### Steps to reproduce

1. Narrow your terminal to < 80 columns (split pane, small window, CI log viewer, etc.)
2. Run `nest g --help`

### Expected behavior

A simple plain list (one schematic per block)

<img width="706" height="909" alt="Image" src="https://github.com/user-attachments/assets/03210b7b-5485-4181-aa53-52f620c58528" />

### Package version

12.0.0

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
lets track this here https://github.com/nestjs/nest-cli/pull/3548

PR #3548, opened 2026-09-04
fix: show schematics as plain list on narrow terminals

biisal:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #3547 

`nest g --help` always renders the schematics list as a `cli-table3` table, even when the terminal with smaller columns. This causes the output to overflow and become unreadable.

image

## What is the new behavior?

When the terminal is narrower than 80 columns (`process.stdout.columns < 80`), the schematics list is rendered as a plain list instead of a table, keeping the output readable. On wider terminals (>= 80 columns), the table is kept with `wordWrap: true` for cleaner wrapping of long descriptions.

image

```
[ ] Yes
[x] No
```

No API change — this only affects help-text rendering in the CLI.

## Other information

- **Changed file:** `commands/generate.command.ts` (1 file, +17 / -2 lines)
- **Tested by:** resizing terminal to < 80 columns and running `nest g --help`
- Tests not added since this is a minor UI/output change in the help text renderer.

### DIFF
--- a/commands/generate.command.ts
+++ b/commands/generate.command.ts
@@ -1,5 +1,5 @@
 import { bold, cyan, green } from 'ansis';
-import Table from 'cli-table3';
+import Table, { TableConstructorOptions } from 'cli-table3';
 import { Command } from 'commander';
 import {
   AbstractCollection,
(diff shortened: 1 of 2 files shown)


micalevisk:
What about using the lib `tty-table` instead?

biisal:
Yup, I tried tty-table as well, but from what I found, it only wraps the existing columns at narrower widths. Names and aliases quickly get split across multiple lines in smaller width (and its not even too much smaller), which creates the same issue.
Since I couldn't see any layout switching feature, I stayed with cli-table3 and the narrow-terminal fallback

PR #2627, opened 2024-07-03
fix: propagate errors raised when loading the webpack configuration file

micalevisk:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: closes #2421

## What is the new behavior?

propagate any errors raised when loading the webpack configuration file. For example:

<details>
<summary>1. when webpack config path is not supplied but the file do exists</summary>

in this case I didn't installed `run-script-webpack-plugin` and the config file is using it

image

</details>

<details>
<summary>2. when the supplied webpack config path is not found</summary>

image

</details>

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No(?)
```

PR #3553, opened 2026-09-05
feat(compiler): support native ESM for bundler config files

### LINKED ISSUE #3552: [Feature] Support native ESM (export default) for Rspack and Webpack config files via dynamic import()

In NestJS v12, applications are increasingly configured with native ESM (`"type": "module"` in `package.json`) and Rspack as the default compiler.

However, when configuring a custom bundler config (e.g. `rspack.config.js` or `webpack.config.js`) via `compilerOptions.builder.options.configPath` in `nest-cli.json`:

```javascript
// rspack.config.js
export default function (options) {
  return {
    ...options,
  };
}
```

This fails or behaves unexpectedly depending on the Node.js version:

1. **On Node < 22 (or without `--experimental-require-module`)**:
   `nest build` crashes with:
   ```text
   Error [ERR_REQUIRE_ESM]: require() of ES Module rspack.config.js not supported.
   ```

2. **On Node 22+ / Node 24 (Silent Bug)**:
   Node allows `require(esm)`, but it returns a module namespace object:
   ```javascript
   [Module: null prototype] { __esModule: true, default: [Function: default] }
   ```
   In `rspack-compiler.ts` (and `webpack-compiler.ts`), `@nestjs/cli` checks:
   ```typescript
   const projectRspackOptions = typeof extras.rspackConfigFactoryOrConfig !== 'function'
       ? extras.rspackConfigFactoryOrConfig
       : extras.rspackConfigFactoryOrConfig(defaultOptions, rspack);
   ```
   Because `typeof` the module namespace is `'object'` (not `'function'`), `@nestjs/cli` **silently ignores the user's config function and never invokes it**, spreading `{ __esModule: true, default: [Function] }` into the options without warning.

Developers are currently forced to maintain `.cjs` files (`rspack.config.cjs`) using legacy `module.exports` syntax even in pure ESM codebases.

In `actions/build.action.ts`, convert `getRspackConfigFactoryByPath` and `getWebpackConfigFactoryByPath` to asynchronous methods using Node's native `import()`:

```typescript
import { pathToFileURL } from 'node:url';

private async getRspackConfigFactoryByPath(
  rspackPath: string,
  defaultPath: string,
): Promise<(config: Record<string, any>, rspackRef: any) => Record<string, any>> {
  const pathToRspackFile = join(process.cwd(), rspackPath);
  const isRspackFileAvailable = isModuleAvailable(pathToRspackFile);
  if (!isRspackFileAvailable && rspackPath === defaultPath) {
    return (_config: Record<string, any>) => ({});
  }

  const fileUrl = pathToFileURL(pathToRspackFile).href;
  const imported = await import(fileUrl);
  return imported.default ?? imported;
}
```

And in `runRspack` / `runWebpack`:
```typescript
const rspackConfigFactoryOrConfig = await this.getRspackConfigFactoryByPath(rspackPath, defaultRspackConfigFilename);
```

Because `import()` natively supports loading CommonJS modules in Node.js, `imported.default ?? imported` resolves both:
- `export default function(options) { ... }` (native ESM)
- `module.exports = function(options) { ... }` (legacy CommonJS)

- **First-class ESM DX**: Allows developers in `"type": "module"` repositories to write standard `export default` in `rspack.config.js` or `webpack.config.js` instead of being forced into `.cjs`.
- **Fixes Silent Bug**: Fixes the issue on Node 22+ where `export default` config functions are silently discarded.
- **Zero Breaking Changes**: Existing CommonJS projects continue working without modification.
- **Zero Dependencies**: Relies exclusively on Node's built-in `node:url` and `import()`.
- **Performance**: Benchmarked overhead between `require()` and `import()` is only ~0.04 ms (40 microseconds).

No migration required. Users can now simply write `export default function (options) { ... }` in `rspack.config.js` or `webpack.config.js`.

I would be happy to submit a Pull Request with the fix and unit tests!

AnasX7:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Resolves #3552

Currently, `getRspackConfigFactoryByPath` and `getWebpackConfigFactoryByPath` in `actions/build.action.ts` load custom configuration files using synchronous `require()`.

When building in an ESM environment (such as a project with `"type": "module"` in `package.json`), attempts to use ESM syntax (`export default`) in `rspack.config.js` or `webpack.config.js` fail:
- On older Node runtimes, `require()` throws `ERR_REQUIRE_ESM`.
- On Node 22+, Node's experimental synchronous `require(esm)` returns an ES Module namespace object (`{ __esModule: true, default: [Function] }`). The CLI checks `typeof config !== 'function'` and treats it as a static options object, silently ignoring the user config factory function.

## What is the new behavior?
1. Both `getRspackConfigFactoryByPath` and `getWebpackConfigFactoryByPath` are now `async` and use native dynamic `await import(pathToFileURL(path).href)`.
2. Automatically unwraps `imported.default ?? imported` to support both ESM (`export default`) and CommonJS (`module.exports`).
3. Callers `runRspack` and `runWebpack` await the factory resolution.
4. Removes unused `createRequire(import.meta.url)`.
5. Adds unit tests verifying both ESM and CommonJS configuration loading and function extraction.

```
[ ] Yes
[x] No
```
Existing CommonJS configuration files continue to work seamlessly, while ESM configuration files are now fully supported.

### DIFF
--- a/actions/build.action.ts
+++ b/actions/build.action.ts
@@ -1,6 +1,6 @@
-import { createRequire } from 'module';
 import { red } from 'ansis';
 import { join } from 'path';
+import { pathToFileURL } from 'url';
 import { BuildCommandContext } from '../commands/index.js';
 import { AssetsManager } from '../lib/compiler/assets-manager.js';
 import { deleteOutDirIfEnabled } from '../lib/compiler/helpers/delete-out-dir.js';
@@ -31,8 +31,6 @@ import { ERROR_PREFIX, INFO_PREFIX } from '../lib/ui/index.js';
 import { isModuleAvailable } from '../lib/utils/is-module-available.js';
 import { AbstractAction } from './abstract.action.js';
 import type webpack from 'webpack';
-
-const require = createRequire(import.meta.url);
 
 export class BuildAction extends AbstractAction {
   protected readonly pluginsLoader = new PluginsLoader();
@@ -308,10 +306,11 @@ export class BuildAction extends AbstractAction {
       getWebpackConfigPath(configuration, options, appName) ??
       defaultWebpackConfigFilename;
 
-    const webpackConfigFactoryOrConfig = this.getWebpackConfigFactoryByPath(
-      webpackPath,
-      defaultWebpackConfigFilename,
-    );
+    const webpackConfigFactoryOrConfig =
+      await this.getWebpackConfigFactoryByPath(
+        webpackPath,
+        defaultWebpackConfigFilename,
+      );
 
     return webpackCompiler.run(
       configuration,
@@ -373,19 +372,22 @@ export class BuildAction extends AbstractAction {
     }
   }
 
-  private getWebpackConfigFactoryByPath(
+  private async getWebpackConfigFactoryByPath(
     webpackPath: string,
     defaultPath: string,
-  ): (
-    config: webpack.Configuration,
-    webpackRef: typeof webpack,
-  ) => webpack.Configuration {
+  ): Promise<
+    (
+      config: webpack.Configuration,
+      webpackRef: typeof webpack,
+    ) => webpack.Configuration
+  > {
     const pathToWebpackFile = join(process.cwd(), webpackPath);
     const isWebpackFileAvailable = isModuleAvailable(pathToWebpackFile);
     if (!isWebpackFileAvailable && webpackPath === defaultPath) {
       return (_config: webpack.Configuration) => ({});
     }
-    return require(pathToWebpackFile);
+    const imported = await import(pathToFileURL(pathToWebpackFile).href);
+    return imported.default ?? imported;
   }
 
   private async runRspack(
@@ -407,7 +409,7 @@ export class BuildAction extends AbstractAction {
       getRspackConfigPath(configuration, options, appName) ??
       defaultRspackConfigFilename;
 
-    const rspackConfigFactoryOrConfig = this.getRspackConfigFactoryByPath(
+    const rspackConfigFactoryOrConfig = await this.getRspackConfigFactoryByPath(
       rspackPath,
       defaultRspackConfigFilename,
     );
@@ -428,16 +430,19 @@ export class BuildAction extends AbstractAction {
     );
   }
 
-  private getRspackConfigFactoryByPath(
+  private async getRspackConfigFactoryByPath(
     rspackPath: string,
     defaultPath: string,
-  ): (config: Record<string, any>, rspackRef: any) => Record<string, any> {
+  ): Promise<
+    (config: Record<string, any>, rspackRef: any) => Record<string, any>
+  > {
     const pathToRspackFile = join(process.cwd(), rspackPath);
     const isRspackFileAvailable = isModuleAvailable(pathToRspackFile);
     if (!isRspackFileAvailable && rspackPath === defaultPath) {
       return (_config: Record<string, any>) => ({});
     }
-    return require(pathToRspackFile);
+    const imported = await import(pathToFileURL(pathToRspackFile).href);
+    return imported.default ?? imported;
   }
 
   private warnOnIgnoredLibraryAssets(
--- a/test/actions/build.action.spec.ts
+++ b/test/actions/build.action.spec.ts
@@ -1,3 +1,4 @@
+import fs from 'fs';
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { BuildAction } from '../../actions/build.action.js';
 import { Configuration } from '../../lib/configuration/index.js';
@@ -84,7 +85,61 @@ describe('BuildAction - Rspack', () => {
   });
 
   describe('getRspackConfigFactoryByPath', () => {
-    it('should return identity function when config file is not available and path is default', () => {
+    it('should load and unwrap export default from ESM config file', async () => {
+      const proto = Object.getPrototypeOf(buildAction);
+      const method =
+        proto.getRspackConfigFactoryByPath ||
+        (buildAction as any)['getRspackConfigFactoryByPath'];
+
+      if (method) {
+        const fixturePath = 'test/fixtures/rspack.esm.config.js';
+        fs.mkdirSync('test/fixtures', { recursive: true });
+        fs.writeFileSync(
+          fixturePath,
+          'export default function (options) { return { ...options, loadedFromEsm: true }; }',
+        );
+
+        const result = await method.call(
+          buildAction,
+          fixturePath,
+          'rspack.config.js',
+        );
+
+        fs.unlinkSync(fixturePath);
+
+        expect(typeof result).toBe('function');
+        expect(result({})).toEqual({ loadedFromEsm: true });
+      }
+    });
+
+    it('should load and unwrap module.exports from CommonJS config file', async () => {
+      const proto = Object.getPrototypeOf(buildAction);
+      const method =
+        proto.getRspackConfigFactoryByPath ||
+        (buildAction as any)['getRspackConfigFactoryByPath'];
+
+      if (method) {
+        const fixturePath = 'test/fixtures/rspack.cjs.config.cjs';
+        fs.mkdirSync('test/fixtures', { recursive: true });
+        fs.writeFileSync(
+          fixturePath,
+          'module.exports = function (options) { return { ...options, loadedFromCjs: true }; }',
+        );
+
+        const result = await method.call(
+          buildAction,
+          fixturePath,
+          'rspack.config.js',
+        );
+
+        fs.unlinkSync(fixturePath);
+
+        expect(typeof result).toBe('function');
+        expect(result({})).toEqual({ loadedFromCjs: true });
+      }
+    });
+
+    it('should return identity function when config file is not available and path is default', async () => {
       // Access private method via prototype
       const proto = Object.getPrototypeOf(buildAction);
       const method =
@@ -93,7 +148,7 @@ describe('BuildAction - Rspack', () => {
 
       // If method exists on prototype, call it bound
       if (method) {
-        const result = method.call(
+        const result = await method.call(
           buildAction,
           'rspack.config.js',
           'rspack.config.js',


AnasX7:
Noticed that PR #3535 was also recently opened for Rspack!

This PR expands on that by:
1. Symmetrically fixing **both** Rspack and Webpack (`webpack.config.js` had the exact same issue).
2. Ensuring full Windows compatibility by wrapping paths with `pathToFileURL(path).href` (in Node ESM, absolute Windows paths like `C:\\path` throw `ERR_UNSUPPORTED_ESM_URL_SCHEME` if imported directly as raw strings).
3. Allowing native dynamic `import()` to load both ESM and CommonJS universally without needing manual extension checks (e.g. `.endsWith('.cjs')`).

actions/build.action.ts
```diff
@@ -373,19 +372,22 @@ export class BuildAction extends AbstractAction {
...
+      webpackRef: typeof webpack,
+    ) => webpack.Configuration
+  > {
     const pathToWebpackFile = join(process.cwd(), webpackPath);
     const isWebpackFileAvailable = isModuleAvailable(pathToWebpackFile);
     if (!isWebpackFileAvailable && webpackPath === defaultPath) {
       return (_config: webpack.Configuration) => ({});
     }
-    return require(pathToWebpackFile);
+    const imported = await import(pathToFileURL(pathToWebpackFile).href);
```
micalevisk: Regression (verified on Node 22 and 26): `require()` resolved extensionless paths, directory `index.js` and `.json`; a `file://` URL passed to `import()` does none of that, while `isModuleAvailable()` two lines above still uses `require.resolve` and says the file exists. `configPath: "config/rspack"` (either `config/rspack.js` or `config/rspack/index.js`) and `rspack.config.json` all work on `master` and crash on this branch.

Suggested shape (prototyped, passes all cases):

```ts
const resolved = require.resolve(pathToWebpackFile);
if (extname(resolved) === '.json') return require(resolved);
const mod = await import(pathToFileURL(resolved).href);
return mod.default ?? mod;
```

with `require = createRequire(import.meta.url)` kept (or a `resolveModulePath()` sibling of `isModuleAvailable` so both use one resolver).

actions/build.action.ts
```diff
@@ -373,19 +372,22 @@ export class BuildAction extends AbstractAction {
...
+    return imported.default ?? imported;
```
micalevisk: Non-blocking, fine as a follow-up: a config compiled by TypeScript to CJS (`exports.__esModule = true; exports.default = fn`) comes through `import()` as `default = module.exports`, i.e. the wrapper object, so this returns the object and the factory is still silently ignored (same class as #3552; also true on master). webpack-cli handles it with `mod.__esModule && 'default' in mod ? mod.default : mod` after the first unwrap.

actions/build.action.ts
```diff
@@ -428,16 +430,19 @@ export class BuildAction extends AbstractAction {
...
+  ): Promise<
+    (config: Record<string, any>, rspackRef: any) => Record<string, any>
+  > {
     const pathToRspackFile = join(process.cwd(), rspackPath);
     const isRspackFileAvailable = isModuleAvailable(pathToRspackFile);
     if (!isRspackFileAvailable && rspackPath === defaultPath) {
       return (_config: Record<string, any>) => ({});
     }
-    return require(pathToRspackFile);
+    const imported = await import(pathToFileURL(pathToRspackFile).href);
```
micalevisk: Same regression as in `getWebpackConfigFactoryByPath`. These two methods are now identical apart from the type annotation; please extract one shared helper under `lib/utils/` (e.g. next to `local-binaries.ts`, which already uses `import(pathToFileURL(...).href)`) and test that helper directly instead of the private methods.

test/actions/build.action.spec.ts
```diff
@@ -84,7 +85,61 @@ describe('BuildAction - Rspack', () => {
...
 
   describe('getRspackConfigFactoryByPath', () => {
-    it('should return identity function when config file is not available and path is default', () => {
+    it('should load and unwrap export default from ESM config file', async () => {
+      const proto = Object.getPrototypeOf(buildAction);
+      const method =
+        proto.getRspackConfigFactoryByPath ||
+        (buildAction as any)['getRspackConfigFactoryByPath'];
+
+      if (method) {
```
micalevisk: This guard makes the test vacuous: if the private method is renamed, the test passes with zero assertions. The pre-existing test below has the same pattern, but new tests should not inherit it. Testing a shared loader helper directly (see comment in `build.action.ts`) removes the need for the prototype lookup altogether.

test/actions/build.action.spec.ts
```diff
@@ -84,7 +85,61 @@ describe('BuildAction - Rspack', () => {
...
+        const fixturePath = 'test/fixtures/rspack.esm.config.js';
+        fs.mkdirSync('test/fixtures', { recursive: true });
```
micalevisk: Fixtures are written into the repo tree under `process.cwd()`: an empty `test/fixtures/` directory is left behind after every run (checked), the file leaks if `import()` throws (the `unlinkSync` runs after the call), and parallel vitest workers writing the same fixed path would race. Other specs here (`test/lib/utils/mau.spec.ts`, `test/lib/compiler/helpers/path-confinement.spec.ts`) use `mkdtempSync(join(tmpdir(), 'nest-...'))` in `beforeEach` and `rmSync(dir, { recursive: true, force: true })` in `afterEach`; please do the same. A temp dir per test also sidesteps Node's ESM cache being keyed by URL.

test/actions/build.action.spec.ts
```diff
@@ -84,7 +85,61 @@ describe('BuildAction - Rspack', () => {
...
   describe('getRspackConfigFactoryByPath', () => {
-    it('should return identity function when config file is not available and path is default', () => {
+    it('should load and unwrap export default from ESM config file', async () => {
+      const proto = Object.getPrototypeOf(buildAction);
+      const method =
+        proto.getRspackConfigFactoryByPath ||
+        (buildAction as any)['getRspackConfigFactoryByPath'];
+
+      if (method) {
+        const fixturePath = 'test/fixtures/rspack.esm.config.js';
```
micalevisk: Nit: this `.js` file is only parsed as ESM because this repo's own `package.json` is `"type": "module"`. Use `.mjs` so the fixture is self-describing and the test does not depend on the CLI's packaging.

test/actions/build.action.spec.ts
```diff
@@ -84,7 +85,61 @@ describe('BuildAction - Rspack', () => {
...
+        );
+
+        fs.unlinkSync(fixturePath);
+
+        expect(typeof result).toBe('function');
+        expect(result({})).toEqual({ loadedFromEsm: true });
+      }
+    });
+
+    it('should load and unwrap module.exports from CommonJS config file', async () => {
```
micalevisk: Coverage gaps: there is no test for `getWebpackConfigFactoryByPath` (identical code, but the PR claims symmetric support), none for the plain-object / array shapes the compilers accept (`RspackConfigFactoryOrConfig[]`), and none for the extensionless / directory-index / `.json` `configPath` regressions once fixed.

micalevisk (changes requested):
Thanks for the PR, and for consolidating #3535 into a single change for both bundlers. The direction is right (Copilot flagged exactly this `require()` problem on the v12 release PR, #3280, and nobody followed up) and the headline case is fixed. I ran this branch against real projects (rspack builder, `nest build`) on Node 22.22 and 26.3 and compared with `master`, so the notes below are from observed behaviour, not reading.

### What I verified works

| scenario | master | this PR |
|---|---|---|
| `"type": "module"` + `rspack.config.js` with `export default fn` (#3552) | factory silently ignored | ✅ factory invoked |
| `export default [ {...}, fn ]` (multi-config array) | silently ignored | ✅ |
| `export default { ... }` plain object | silently ignored | ✅ |
| ESM named exports only (`export const output = ...`) | ✅ | ✅ |
| `rspack.config.cjs` via `configPath` in an ESM project | ✅ | ✅ |
| CJS project + `module.exports = fn` | ✅ | ✅ |
| CJS project + `webpack.config.mjs` via `--webpackPath` | crashes ("Invalid configuration object") | ✅ factory invoked |
| missing default config file → identity factory (#2627 behaviour) | ✅ | ✅ |

Full unit suite (996 tests), `tsc`, `oxlint` and `prettier` are green on the branch.

### Blocking

**1. `import()` does not resolve like `require()` did, so three previously working `configPath` shapes now crash.** `require()` applied CommonJS resolution (extensionless, directory `index.js`, `.json`); a `file://` URL passed to `import()` gets none of that. `isModuleAvailable()` still uses `require.resolve`, so the guard says "available" and then the import throws:

| `configPath` | master | this PR |
|---|---|---|
| `config/rspack` → `config/rspack.js` | ✅ | ❌ `Cannot find module` |
| `config/rspack` → `config/rspack/index.js` | ✅ | ❌ `Directory import ... is not supported` |
| `rspack.config.json` | ✅ | ❌ `needs an import attribute of "type: json"` |

The PR is labelled non-breaking, so these need to keep working. The fix is small: resolve the path with the same resolver the guard already uses, then import the resolved file. I prototyped this and it passes all seven shapes above plus the three regressions:

```ts
const resolved = require.resolve(pathToConfigFile); // extensionless / dir index / .json, same as isModuleAvailable
if (extname(resolved) === '.json') return require(resolved);
const mod = await import(pathToFileURL(resolved).href);
return mod.default ?? mod;
```

(`require` here being `createRequire(import.meta.url)`, which this PR removed; keeping it is fine, `isModuleAvailable` could also grow a `resolveModulePath()` sibling so both places share one resolver.)

**2. Extract one shared loader instead of two identical bodies.** `getWebpackConfigFactoryByPath` and `getRspackConfigFactoryByPath` are now byte-for-byte the same apart from the type annotation. Please move the load-and-unwrap logic into a single helper under `lib/utils/` (e.g. next to `local-binaries.ts`, which already uses the `import(pathToFileURL(...).href)` pattern) and unit test that helper directly. That also removes the need to reach private methods through `Object.getPrototypeOf` in the spec.

### Tests

**3. Fixtures are written into the repo tree at runtime.** `test/fixtures/` is created under `process.cwd()` and never removed (an empty `test/fixtures/` directory is left behind after every run, I checked), the file leaks if `import()` throws (the `unlinkSync` is after the call), and two vitest workers writing the same fixed path would race. Other specs in this repo (`test/lib/utils/mau.spec.ts`, `test/lib/compiler/helpers/path-confinement.spec.ts`) use `mkdtempSync(join(tmpdir(), ...))` in `beforeEach` and `rmSync` in `afterEach`; please follow that. Note also that Node's ESM cache is keyed by URL, so a fixed path with different content across tests would return stale modules; a temp dir per test avoids that too.

**4. The `if (method) { ... }` guard makes the new tests vacuous.** If the private method is renamed the tests pass with zero assertions. The pre-existing test has the same pattern, but new tests should not inherit it. With the helper from point 2 this goes away entirely.

**5. Coverage gaps.** There is no test for `getWebpackConfigFactoryByPath` (identical code, but the PR claims symmetric support), none for the plain-object and array shapes the compilers accept (`RspackConfigFactoryOrConfig[]`), and none for the three regressions in point 1 once fixed. The ESM fixture uses `.js` and only works as ESM because this repo's own `package.json` is `"type": "module"`; use `.mjs` so the test is self-contained.

### Non-blocking, can be a follow-up

**6. TypeScript-compiled configs are still silently ignored.** A `webpack.config.ts`/`rspack.config.ts` compiled to CJS emits `exports.__esModule = true; exports.default = fn`. `import()` of that gives `default = module.exports` (the wrapper object), so `imported.default ?? imported` returns the object and the factory is never called; same class of bug as #3552, present on master too. webpack-cli unwraps this shape (`mod.__esModule && 'default' in mod ? mod.default : mod`). Happy to take it as a follow-up if you prefer to keep this PR focused.

**7. Return type.** The methods still declare a factory-only return type while the runtime value can be an object or an array. Pre-existing, but since the signature is being rewritten anyway, mirroring `RspackCompilerExtras['rspackConfigFactoryOrConfig']` / `WebpackCompilerExtras['webpackConfigFactoryOrConfig']` would make the `typeof !== 'function'` branch in the compilers honest.

**8. Docs.** The "Docs" checkbox is unchecked. After this lands, the CLI docs and the `hot-reload` recipe (which currently points people at a `.cjs` config, see nest sample `36-hmr-esm`) can say `export default` is supported. Separate PR to `nestjs/docs.nestjs.com` is fine.

### Context worth knowing

- The webpack compiler already refuses `"type": "module"` projects (`lib/compiler/webpack-compiler.ts:90`), so the webpack half of this change is only reachable with an `.mjs` config in a CJS project. That case is real and now works (row 7 above), but the description overstates it a bit.
- No maintainer has ever declined ESM config support; the prior "resolution" was the `.cjs` workaround in the HMR sample. #3535 was closed by its author in favour of this PR. So nothing here contradicts past decisions.

Requesting changes for points 1 to 5. Once the resolver is back and the tests live in a temp dir, this is good to go.

AnasX7:
Hi @micalevisk, thank you for the detailed and helpful review!

All feedback points have been addressed in the latest commit:

1. **Shared Loader (`lib/utils/load-builder-config.ts`)**: Extracted config loading logic so it is shared across both Webpack and Rspack without duplication.
2. **Resolution & Extension Handling**: 
   - Uses `require.resolve` to preserve Node's classic resolution for extensionless paths (e.g., `config/rspack`) and directory indexes (`config/rspack/index.js`).
   - Loads `.json` configs directly via `require(resolved)`.
   - Converts the resolved path to a file URL (`pathToFileURL(resolved).href`) before dynamic `import()`.
3. **TypeScript CJS Interop & Return Types**:
   - Added unwrapping for TS-compiled CJS modules (`exports.__esModule = true; exports.default = ...`).
   - Aligned method return types with `WebpackCompilerExtras['webpackConfigFactoryOrConfig']` and `RspackCompilerExtras['rspackConfigFactoryOrConfig']`.
4. **Isolated Temp Fixtures & Comprehensive Tests**:
   - Cleaned up `test/actions/build.action.spec.ts`: removed repo-tree file writes and removed the `if (method)` conditional guard so assertions run directly.
   - Added a dedicated test suite in `test/lib/utils/load-builder-config.spec.ts` using `mkdtempSync(join(tmpdir(), 'nest-builder-config-'))` with automatic teardown in `afterEach`.
   - Added full test coverage for:
     - ESM (`.mjs`) functions, plain objects, multi-compiler arrays, and named exports
     - CommonJS (`.cjs`) functions, objects, and TS-compiled `__esModule` interop
     - `.json` configs
     - Extensionless paths and directory index resolution
     - Missing default config (fallback to identity) vs. missing custom config (re-throws error)

All 87 test suites (1,008 tests), linting, and build pass cleanly.

Issue #3479, opened 2026-07-11
Support TypeScript 7: adapt CLI to work without the programmatic Compiler API


joaosousafranco:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Why this issue exists alongside #3477 / #3478

[#3477](https://github.com/nestjs/nest-cli/issues/3477) reported the crash when using `@nestjs/cli` with TypeScript 7.0.2, and [#3478](https://github.com/nestjs/nest-cli/pull/3478) addressed it by adding an early detection guard with an actionable error message. That fix is valuable — users now get a clear "install TypeScript 6" message instead of a raw `TypeError`.

However, #3478 is a diagnostic improvement, not a compatibility fix. The CLI still cannot run `nest build`, `nest start`, or `nest start --watch` with TypeScript 7. This issue tracks the actual work needed to make the CLI functional with TypeScript 7.

### Current behavior

With `typescript@7.0.2` installed, all compiler-dependent CLI commands fail because the Node module no longer exports the programmatic Compiler API (`getParsedCommandLineOfConfigFile`, `createProgram`, `createWatchProgram`, `getPreEmitDiagnostics`, etc.). This affects both the default `tsc` builder and the SWC builder (which uses the programmatic API for type-checking).

The only workaround today is to bypass the CLI entirely for compilation:

```json
{
  "build": "tsc",
  "dev": "concurrently \"tsc --watch --preserveWatchOutput\" \"node --watch-path=dist dist/src/main.js\""
}
```

This works, but users lose CLI features like plugin metadata generation, the integrated SWC type-checker, and the unified `nest build`/`nest start` interface.

### Expected behavior

`nest build` and `nest start` should work with TypeScript 7 once the programmatic API stabilizes (expected in TypeScript 7.1), or the CLI should provide an alternative compilation path that does not depend on it.

### Context

TypeScript 7.0 was rewritten in Go and only ships the `tsc` CLI executable. Microsoft has stated the programmatic Compiler API is expected to return in TypeScript 7.1, but there is no confirmed timeline. As TypeScript 7 adoption grows, more users will hit this incompatibility.

### Package version

11.0.24

### NestJS version

11.1.28

### Node.js version

v26.5.0

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
> This issue tracks the actual work needed to make the CLI functional with TypeScript 7.

[...]

PR #3554, opened 2026-09-05
feat(compiler): support TypeScript 7.1 native API in tsc builder

BeamNawapat:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #3479

With TypeScript 7 installed, `nest build` stops at the guard added in #3478, because `require('typescript')` only exports `version`. Since TypeScript 7.1 (`typescript@next`, `7.1.0-dev.20260905.1` at the time of writing) the package ships a programmatic API under `typescript/unstable/sync` that covers what the `tsc` builder needs: config parsing, program creation, diagnostics and emit (microsoft/typescript-go#4699). Design was proposed in https://github.com/nestjs/nest-cli/issues/3479#issuecomment-5554398504.

**Scope:** `nest build` (one-shot) only. Watch mode is a separate follow-up PR, per review; under TypeScript 7.1 `--watch` currently fails with a dedicated message.

## What is the new behavior?

The `tsc` builder compiles through the native API when the installed TypeScript has no classic compiler API but exposes a complete `typescript/unstable/sync`.

- `TypeScriptBinaryLoader` gains `hasProgrammaticApi()`, `hasNativeApi()`, `isNativeApiRequired()`, `loadNativeModule()`, `getNativeModule()`, `loadNativeApi()` and `closeNativeApi()`. Detection is a feature probe, never a semver comparison: TypeScript 7.0.2 already exports `unstable/sync` but its `API` has `parseConfigFile` only, so the probe requires `API.prototype.createProgram` and `formatDiagnosticsWithColorAndContext`. Results are memoized. An entry point that exists but fails to load is reported as such rather than as "install 7.1".
- `TsConfigProvider` parses through `api.parseConfigFile` on the native path and returns the same shape as before plus `configFileParsingDiagnostics`, which the native `createProgram` needs in order to report tsconfig errors (TS5025 and friends); `build.action.ts` (`outDir`, `rootDir`, `deleteOutDir`, assets) is unchanged.
- New `NativeCompiler` mirrors `Compiler`: `createProgram` -> `emit()` -> diagnostics (global, which is the union of config and options diagnostics; syntactic; bind; semantic; declaration when declarations are emitted; emit) -> exit 1 when any has `DiagnosticCategory.Error`, otherwise `onSuccess()`. Warnings are printed but do not fail the build, matching the classic path. The program and the API session are released on every path, including the ones that throw before a program exists, and `runBuild` closes a session that `TsConfigProvider` opened for a non-`tsc` builder.
- The native `Program.emit()` accepts no custom transformers, so `compilerOptions.plugins` (`@nestjs/swagger`, `@nestjs/graphql`, ...) and `compilerOptions.paths` (rewritten today by `tsconfigPathsBeforeHookFactory`) both fail with an actionable error instead of producing output that breaks at runtime.
- Builders still bound to the classic API (`swc` with `typeCheck`, `webpack`/`rspack`, plugin metadata generation) get a dedicated message under TypeScript 7.1 (`CLASSIC_API_REQUIRED_ON_NATIVE_TYPESCRIPT`) naming the `tsc` builder as the supported path. The TypeScript 7.0 (no usable API) message is updated to point at 7.1. The TypeScript 6 path is untouched.

### Verification

Sample Nest app built with `bin/nest.js` on Node 24.20:

| TypeScript | `nest build` | Result |
|---|---|---|
| 7.1.0-dev.20260905.1 | ok | `dist/main.js` runs, prints from the injected service |
| 7.1 + `compilerOptions.plugins` | error | "Compiler plugins (...) run as TypeScript transformers ..." |
| 7.1 + `compilerOptions.paths` | error | "compilerOptions.paths aliases (...) are rewritten by a TypeScript transformer ..." |
| 7.1 + `--watch` | error | "Watch mode is not supported with the native TypeScript 7 compiler yet ..." |
| 7.1 + `--builder swc --type-check` | error | "... only exposes the native compiler API. The tsc builder supports it, but ..." |
| 7.1 + invalid tsconfig key | exit 1 | TS5025 rendered |
| 7.1 + type error | exit 1 | TS2322 rendered |
| 7.0.2 | error | "... does not expose a programmatic compiler API ... install TypeScript 7.1 or newer, or TypeScript 6" |
| 6.0.3 | ok | unchanged behaviour |

Tests: `npm test` 1016 pass. New: `test/lib/compiler/native/native-compiler.spec.ts`; detection tests in `typescript-loader.spec.ts` against fake `typescript` packages on disk shaped like 6.0, 7.0.2 (`unstable/sync` with `parseConfigFile` only), 7.1, a broken 7.1 install, and the documented side-by-side layout from #3479 (`"typescript": "npm:@typescript/typescript6@^6"` + `"@typescript/native": "npm:typescript@^7"`). New e2e `test/e2e/typescript-7.e2e-spec.ts` scaffolds an app, installs `typescript@next`, and checks build output, a type error and the plugins error against the real API; it skips itself if the nightly cannot be installed or drops `createProgram` again. `tsc --noEmit` clean.

```
[ ] Yes
[x] No
```

## Other information

- Watch mode (`nest build --watch`, `nest start --watch`) is implemented on a branch and will follow as a second PR once this one lands, with the watch-root, `onUnlink` and `onSuccess` findings from the review addressed there.
- Plugin and `paths` support depend on the native emit exposing transformers upstream; until then the side-by-side layout from #3479 is the documented way to keep them.
- A stock `nest new` project needs `rootDir` set (TS5011) and `baseUrl` removed for TypeScript 7; `@nestjs/schematics` defaults and docs are not touched here.
- The native API is still marked unstable (microsoft/TypeScript#64154). TypeScript 7.1 beta is scheduled for 2026-10-06 and stable for 2026-11-24 (microsoft/TypeScript#63703).


micalevisk (commented):
Reviewed `ab0b5d4` against a real `typescript@next` (`7.1.0-dev.20260905.1`), probing
`typescript/unstable/sync` directly — everything marked **verified** was reproduced, not
reasoned about. Confirmed your claims: 88 files / 1021 tests pass, `tsc --noEmit` clean.

## Recommendation

1. **Take your own offer and split the watch mode into a second PR.** 417 of ~1700 new lines
   are `native-watch-compiler.ts`, and it holds every blocking finding below.
2. **Get an explicit yes from @kamilmysliwiec on the strategy first** — his only recorded
   position is *"I'd rather wait for the programmatic API to become stable"*, and the PR
   doesn't answer that head-on. Don't leave 1700 lines waiting on a one-sentence decision.
3. `paths` as a **warning** ships JS with unresolved aliases → runtime `MODULE_NOT_FOUND`.
   Make it an error or an opt-in, consistent with how you handle `plugins`.

## Blocking

**1. The watch root is wrong in both directions.** The scope comes from
`getEffectiveRootDir(rootDir, fileNames)`, filtered only by extension / `node_modules` /
`outDir`. **Verified** on a stock Nest layout (`tsconfig.build.json`, no `rootDir`):

| | classic `tsc --watch` | this PR |
|---|---|---|
| scope | `wildcardDirectories = { <projectRoot>: 1 }` | `<projectRoot>/src` |
| filter | tsconfig `include`/`exclude` | file extension |

- *False positives:* `src/**/*.spec.ts` is excluded from the program but inside the watch
  root → every spec save rebuilds and restarts the app under `nest start --watch`.
- *False negatives:* inputs outside the root (`typings/*.d.ts`, monorepo `libs/`) never
  rebuild. `getEffectiveRootDir` also deliberately drops `.d.ts` when computing the common
  dir, so declaration inputs are structurally excluded.

Also an altitude problem: `getEffectiveRootDir` is documented as the effective rootDir *for
an emit*. Suggest watching `dirname(tsconfigPath)` and filtering against the `fileNames` /
`exclude` that `TsConfigProvider` already returns — `exclude` is on `TsConfigProviderOutput`
and unused here.

**2. `onUnlink` lands in the shared watcher with no tests.**
`recursive-directory-watcher.ts` is also used by `swc-compiler.ts` (2 call sites). The change
looks additive, but the spec has **zero** `onUnlink` references — neither the `poll()` nor
the `scan(emit = true)` path. Also: `scan()`'s "directory gone" catch and `rearm()` both
`known.clear()` without emitting, so a removed source dir reports no deletions and the next
scan reports every file as `onAdd`.

**3. `onSuccess` gating diverges from classic watch.** Native fires on
`diagnostics.length === 0`; classic fires on tsc's `"0 errors"` status — errors only.
Suggestions are safely out of scope (**verified**: they live in `getSuggestionDiagnostics()`,
which you correctly never call), but `DiagnosticCategory.Warning === 0` is reachable from the
phases you do call. Gate on `category === Error`; the enum is exported from the module but
your interface file doesn't declare it.

## Reuse instead of reimplementation

**4. `stripJsonComments` + `resolveConfigChain` + `tryResolveModule` (~60 lines).**
**Verified:** `api.readConfigFile(path)` returns `{ config }` with comments and trailing
commas already handled. Deletes the regex stripper outright. (Your comment that
`program.getConfigFileNames()` is empty for a program built from parsed options is
**correct** — it returns `[]`. But the `extends` walk belongs in `TsConfigProvider`, next to
the parse it duplicates.)

**5. A second hand-rolled watcher.** `run()` adds raw `fs.watch(dir)` handles for config
dirs, re-deriving rename-replacement handling and error swallowing that
`watchDirectoryRecursively` already solves (plus rearm-on-delete, catch-up scans,
`awaitWriteFinish`). Reuse it with `extensions: ['.json']`.

**6. `isIgnored()`** duplicates the containment check in `helpers/path-confinement.ts`.
Export an `isPathInside` there instead of a fourth copy.

**7. The emit → collect → format → count block** is duplicated between `NativeCompiler.run`
and `NativeWatchCompiler.compile` — that's what `NativeCompilerBase` is for.

**8. `UNSUPPORTED_TYPESCRIPT_VERSION('unknown')`** — the literal string reaches the user,
while `tryGetInstalledVersion()` (added in this PR, private) already produces the right
value. Or drop the branch: routed through `isNativeApiRequired()`, `loadNativeModule()` can't
be `null` there.

**9. `pendingChanges` should hold `Set<string>`** — `mergeChanges` becomes a set union,
`touchesAny` a `.has()`, and repeated saves stop pushing duplicate paths. ~25 lines → 5.

## Assertness

**10. `loadNativeModule()` memoizes `null` on *any* throw,** including a real load failure
inside the module — a broken install then surfaces as "TypeScript 7.0 ships the executable
only". Distinguish `MODULE_NOT_FOUND`.

**11. Detection re-runs on every rebuild.** `hasProgrammaticApi()` isn't memoized
(`require.resolve` with custom `paths` each call) and `isNativeApiRequired()` is called from
`getByConfigFilename()` — once per rebuild in watch mode.

**12. `dedupeDiagnostics` works, but shouldn't need to exist.** **Verified** with a broken
tsconfig: 5 raw diagnostics dedupe correctly to 3, the shapes do line up. But the reason it's
needed is that **`getGlobalDiagnostics()` already returns the union of
`getConfigFileParsingDiagnostics()` and `getProgramDiagnostics()`** (verified — same TS5011
and TS5025 in all three). Dropping those two phases gives the same result, removes two IPC
round-trips per compile, and removes the helper whose own comment admits it's guessing at the
shape.

**13. `TsConfigProvider` acquires the API session as a side effect.** `build.action.ts:99`
parses for *every* builder, so `swc`/`webpack`/`rspack` open a session they never close. Not
a leak — **verified** the child dies with the parent (~90 ms, no orphan) — but ownership is
upside-down versus the rest of the PR.

**14. Minor lifecycle:** `runNativeTsc`'s `try/finally { closeNativeApi() }` is dead code
(`run()` already closes in its own `finally`); `close()` never removes the `exit`/`SIGINT`
listeners it installed; config watchers are registered **once** from the initial chain, so an
`extends` added later is never watched; `flush()`'s catch re-merges the batch but never
reschedules (matches `tsc --watch` on a broken tsconfig — worth a comment, it reads like an
oversight).

**Credit:** the SIGINT handling is correct and subtler than it looks — `start.action.ts`
registers its handler before `runBuild`, and `process.once` self-removes before
`listenerCount` is read, so `exit(130)` fires only for `nest build --watch`. Worth a comment
so nobody "simplifies" it later.

## Tests

- The detection tests are the right shape — real packages on disk, not mocks.
- **Missing:** the side-by-side layout from #3479 (`"typescript": "npm:@typescript/typescript6@^6"`
  + `"@typescript/native": "npm:typescript@^7"`). It reduces logically to the existing
  "prefers the classic API" case, but it's the documented workaround people will file issues
  about. Ten lines.
- Both native specs are mock-only against the hand-written interface, so real-API shape drift
  keeps the suite green — and two of the shapes I probed are right by luck (the
  `getGlobalDiagnostics` overlap; `projectReferences` absent unless the tsconfig has
  `references`). One `typescript@next` e2e fixture, even allow-failure, would catch that.
- *"does not call onSuccess when the rebuild has errors"* asserts `toHaveBeenCalledTimes(1)`
  where the `1` is the **initial** compile — correct, but `mockClear()` then
  `not.toHaveBeenCalled()` says what it means.

## Not a regression (checked, so you don't have to)

- **Output layout identical** — TS6 and TS7.1 emit to the same paths for the same tsconfig.
- **Diagnostic parity holds** — a stock Nest tsconfig reports TS5011 + TS5102 on 7.1, but TS6
  reports TS5011 + TS5101 for the same file. TypeScript's migration requirement, not yours.
- **`projectReferences` survives** — `api.parseConfigFile` returns them when the tsconfig has
  `references`.
- **`noEmitOnError` is honoured** by the native `emit()` (`emitSkipped: true`, nothing
  written), so emitting before collecting diagnostics is safe.
- **`rs` doesn't orphan a session** — `close()` skips `closeNativeApi()`, but the session is
  memoized on the loader and reused. The PR description's "released on `rs`" is inaccurate;
  fix the wording, not the code.

One gap worth naming: a stock Nest project *does* need tsconfig changes for TS 7 (`rootDir`,
`baseUrl`), and neither `@nestjs/schematics`' default tsconfig nor any docs are updated here.

---

*Review generated with Claude Code Opus 5. Findings marked
**verified** were reproduced locally against `typescript@7.1.0-dev.20260905.1`.*

BeamNawapat:
Thanks for the thorough pass, and for verifying against the real nightly. Taken up as follows (see the updated description and the new commit):

**Split.** Watch mode is out of this PR. `nest build --watch` under TypeScript 7.1 now fails with a dedicated message. The watch implementation lives on a branch and will come back as a second PR with the watch-root (`dirname(tsconfig)` filtered by `fileNames`/`exclude`), `onUnlink` tests, and `category === Error` gating addressed there.

**`paths`** is an error now, same shape as the plugins check.

**Reuse / asserted-ness**, in your numbering: 7 (shared `emitAndReport` in the base), 8 (`getNativeModule()` throws with the installed version), 10 (a native entry point that exists but throws on load surfaces its own error; only a missing export maps to `null`), 11 (installed module and probe result memoized), 12 (`getGlobalDiagnostics()` only, dedupe helper deleted), 13 (`runBuild` closes the session for non-`tsc` builders), 14 (dead `try/finally` removed). 4, 5, 6, 9 were watch-only and go with the second PR.

One correction to the detection story that your probe would have hit next: TypeScript 7.0.2 already exports `typescript/unstable/sync`, with `API.parseConfigFile` but no `createProgram` and no formatter. The probe now requires `createProgram` on `API.prototype`, and the 7.0 test fixture mirrors that shape.

**Tests.** Added the side-by-side layout case, a broken-install case, warnings-not-counted, and an e2e (`test/e2e/typescript-7.e2e-spec.ts`) that scaffolds an app, installs `typescript@next`, and runs `nest build` against the real API; it skips itself if the nightly regresses.

@kamilmysliwiec the open question is the strategy one: this only adds a compiler backend behind a feature probe, the TypeScript 6 path is byte-for-byte unchanged, and the API surface it depends on is the one Microsoft has said 7.1 stabilises (config parse, `createProgram`, diagnostics, `emit`). If you would rather hold until 7.1 beta (2026-10-06) I can keep this rebased until then; if the direction is fine I will open the watch-mode PR on top.

kamilmysliwiec:
lets hold until beta

Issue #3041, opened 2025-05-25
Allow comments in nest-cli.json


mterczynski:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

My problem: I've added "builder": "webpack" option to solve a very specific issue and want to add a comment to that line. Unfortunately nest-cli.json doesn't support comments.

### Describe the solution you'd like

Allow comments in nest-cli.json or use alternative solution to allow comments: for example change json format to json5

### What is the motivation / use case for changing the behavior?

Comment solutions to edge cases

micalevisk:
I'd say that this is not up to `nest-cli.json` as it is supposed to be a valid JSON. We would have to support a file named `nest-cli.jsonc` instead

kamilmysliwiec:
Thank you for taking the time to submit your report! From the looks of it, this could be better discussed on our Discord. If you haven't already, please join [here](https://discord.gg/nestjs) and send a new post in the #⁠ 🐈 nestjs-help forum. Make sure to include a link to this issue, so you don't need to write it all again. We have a large community of helpful members, who will assist you in getting this to work.

Issue #2343, opened 2023-10-20
Ability to override default swc cliOptions when using SWC compiler


pauliusg:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

In my project I have to import package.json file:
```
import { version } from 'package.json';
```
When using SWC  compiler, I get this error: 
```
> nest start --watch

>  SWC  Running...
Successfully compiled: 18 files with swc (41.56ms)
Watching for file changes.
Error: Cannot find module 'package.json'
Require stack:
- c:\app\Grainger\workstation-auth\api\dist\main.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Function.Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:121:18)
    at Object.<anonymous> (c:\app\Grainger\workstation-auth\api\dist\main.js:5:22)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Function.Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
```

This is because SWC is not copying `json` files to `dist` folder. 
To copy those files `--copy-files` cli option has to be used: https://swc.rs/docs/usage/cli#--copy-files--d

Also, I have tried this configuration in `nest-cli.json` file:
```
{
  "$schema": "https://json.schemastore.org/nest-cli",
  "collection": "@nestjs/schematics",
  "sourceRoot": "src",
  "compilerOptions": {
    "builder": "swc",
    "options": {
      "swcrcPath": ".swcrc",
      "copyFiles": true
    },
    "deleteOutDir": true
  }
}
```
But it still does not work.

### Describe the solution you'd like

I see that NestJS swc compiler gets default with `false`:
https://github.com/nestjs/nest-cli/blob/7a271cdba0031e94a45356bb67867f65be63ee5e/lib/compiler/defaults/swc-defaults.ts#L42

We need an option to override default `cliOptions`: 
[...]

PR #3577, opened 2026-09-17
fix(swc): parse .swcrc with a jsonc parser to match swc

stefanwaldhauser:
## Problem

`SwcCompiler.getSwcRcFileContentIfExists` reads `.swcrc` with raw `JSON.parse` and swallows any parse error into `{}` on the default-path branch:

That silently discards the user's config in the very common case of a `.swcrc` that carries a leading `//` comment. Nest's hardcoded `module: { type: "commonjs" }` default in `swc-defaults.ts` then wins, and projects that had configured ESM in `.swcrc` get CJS output with no warning. This trap cost real time to diagnose.

## Why comments are legal in `.swcrc`

SWC parses `.swcrc` with [`jsonc_parser::parse_to_serde_value`](https://github.com/swc-project/swc/blob/476ba3384ed3d1453fae9a1b0d83918a25c3e4fa/crates/swc/src/lib.rs#L132) and explicitly enables both `allow_comments` and `allow_trailing_commas` — [`parse_swcrc`](https://github.com/swc-project/swc/blob/476ba3384ed3d1453fae9a1b0d83918a25c3e4fa/crates/swc/src/lib.rs#L1432-L1465):

```rust
let v = parse_to_serde_value(
    s.trim_start_matches('\u{feff}'),
    &ParseOptions {
        allow_comments: true,
        allow_trailing_commas: true,
        allow_loose_object_property_names: false,
    },
)?
```

So JS-style comments and trailing commas are officially part of the `.swcrc` input format. Nest CLI's `JSON.parse` is stricter than the tool it's meant to configure.

## Fix

- Switch to [`jsonc-parser`](https://www.npmjs.com/package/jsonc-parser) (Microsoft's, zero deps, ~140 KB, same package VS Code uses) with `allowTrailingComma: true` — matches SWC's parser semantics.
- Distinguish *"file not found"* from *"file exists but invalid"*:
  - Default `.swcrc` missing → still returns `{}` (unchanged behavior).
  - `.swcrc` **present** but unparseable → exits with a clear error message and the offset of the failure, regardless of whether the user configured a custom `swcrcPath` or not. Previously the default-path branch swallowed this and returned `{}`.

## Verification

Unit tests added covering:
- Missing default `.swcrc` (returns `{}`, no exit).
- Valid `.swcrc`.
- `.swcrc` with `//` line comments.
- `.swcrc` with `/* */` block comments.
- `.swcrc` with a trailing comma.
- Invalid default `.swcrc` (exits loudly).
- Missing user-specified `swcrcPath` (exits loudly).
- Invalid user-specified `swcrcPath` (exits loudly).

Full existing test suite continues to pass: 87 files, 1,040 tests. `npm run build` and `npm run lint` clean.


lib/compiler/swc/swc-compiler.ts
```diff
@@ -276,20 +281,44 @@ export class SwcCompiler extends BaseCompiler {
...
       return {};
     }
+
+    // .swcrc is JSON-with-comments in SWC itself, which uses jsonc_parser
+    // with `allow_comments` + `allow_trailing_commas` enabled (see swc's
+    // `parse_swcrc` in crates/swc/src/lib.rs). Match that here so we don't
+    // silently reject a valid `.swcrc` that carries `//` / `/* */` comments
+    // or a trailing comma.
+    const errors: ParseError[] = [];
+    const parsed = parseJsonc(content, errors, { allowTrailingComma: true });
```
micalevisk: **Blocking.** A UTF-8 BOM makes this fail with `InvalidSymbol at offset 0`, whereas SWC strips it before parsing (`trim_start_matches('\u{feff}')`). Verified end-to-end: a BOM-prefixed `.swcrc` builds (exit 0) on the parent commit and exits 1 with this change.

```suggestion
    const parsed = parseJsonc(content.replace(/^﻿/, ''), errors, {
      allowTrailingComma: true,
    });
```

I checked that with the BOM stripped, `jsonc-parser` returns the expected object with no errors. Please cover it with a test, e.g. `writeFileSync(join(tmpDir, '.swcrc'), '﻿{ "module": { "type": "es6" } }')`.

lib/compiler/swc/swc-compiler.ts
```diff
@@ -276,20 +281,44 @@ export class SwcCompiler extends BaseCompiler {
...
+    const isCustomPath = swcrcFilePath !== undefined;
+    const resolvedPath = join(process.cwd(), swcrcFilePath ?? '.swcrc');
+
+    let content: string;
     try {
-      return JSON.parse(
-        readFileSync(join(process.cwd(), swcrcFilePath ?? '.swcrc'), 'utf8'),
-      );
+      content = readFileSync(resolvedPath, 'utf8');
     } catch {
```
micalevisk: Non-blocking: the comment says "File missing", but this catches any read error. With the default path, a `.swcrc` that exists but can't be read (`EACCES`, or `EISDIR` — I reproduced the latter) still silently becomes `{}`, which is the behaviour this PR is removing for parse errors. Consider only treating `ENOENT` as "missing":

```ts
} catch (err) {
  const isMissing = (err as NodeJS.ErrnoException)?.code === 'ENOENT';
  if (isCustomPath || !isMissing) {
    // report + process.exit(1)
  }
  return {};
}
```

lib/compiler/swc/swc-compiler.ts
```diff
@@ -276,20 +281,44 @@ export class SwcCompiler extends BaseCompiler {
...
+    // silently reject a valid `.swcrc` that carries `//` / `/* */` comments
+    // or a trailing comma.
+    const errors: ParseError[] = [];
+    const parsed = parseJsonc(content, errors, { allowTrailingComma: true });
+    if (errors.length > 0) {
+      const first = errors[0];
+      console.error(
+        ERROR_PREFIX +
+          ` Failed to parse "${swcrcFilePath ?? '.swcrc'}": ` +
+          `${printParseErrorCode(first.error)} at offset ${first.offset}. ` +
```
micalevisk: Nit: a raw character offset is hard to use on a multi-line file. Line/column is cheap to derive:

```ts
const before = content.slice(0, first.offset).split('\n');
const position = `${before.length}:${before[before.length - 1].length + 1}`;
```

(If you strip the BOM as suggested above, compute this against the same stripped string so the offsets line up.)

lib/compiler/swc/swc-compiler.ts
```diff
@@ -276,20 +281,44 @@ export class SwcCompiler extends BaseCompiler {
...
+          `The file must be valid JSON with comments (JSONC).`,
+      );
+      process.exit(1);
+    }
+    return parsed ?? {};
```
micalevisk, line 329: Follow-up material, not required here: `parsed` can be any JSON value. `42`, `"str"` or SWC's array-of-configs form parse without errors and go straight into `deepMerge`, which then yields a non-object or an options object with a `"0"` key. This was already the case with `JSON.parse`, so I'm fine leaving it for a separate PR.

test/lib/compiler/swc/swc-compiler.spec.ts
```diff
@@ -869,4 +869,101 @@ describe('SWC Compiler', () => {
...
+  describe('getSwcRcFileContentIfExists', () => {
+    let tmpDir: string;
+    let originalCwd: string;
+    let exitSpy: ReturnType<typeof vi.spyOn>;
+    let errorSpy: ReturnType<typeof vi.spyOn>;
+
+    beforeEach(() => {
+      tmpDir = mkdtempSync(join(tmpdir(), 'nest-swcrc-'));
+      originalCwd = process.cwd();
+      process.chdir(tmpDir);
```
micalevisk: `process.chdir()` isn't used anywhere else in the test suite, and it throws inside worker threads (`pool: 'threads'`), so it would couple these tests to Vitest's current default pool. The method only calls `process.cwd()`, so a spy is enough and needs no restore bookkeeping for the cwd:

```ts
vi.spyOn(process, 'cwd').mockReturnValue(tmpDir);
```

I ran the comment/trailing-comma case this way and it passes.

test/lib/compiler/swc/swc-compiler.spec.ts
```diff
@@ -869,4 +869,101 @@ describe('SWC Compiler', () => {
...
+      exitSpy = vi
+        .spyOn(process, 'exit')
+        .mockImplementation(((code?: number) => {
+          throw new Error(`process.exit(${code})`);
+        }) as never);
```
micalevisk: Nit: `prettier --check` fails on this file because of this block (the parent commit passes). `npm run format` rewrites it as:

```suggestion
      exitSpy = vi.spyOn(process, 'exit').mockImplementation(((
        code?: number,
      ) => {
        throw new Error(`process.exit(${code})`);
      }) as never);
```

micalevisk (changes requested):
Thanks for this, @stefanwaldhauser — the diagnosis is right and the fix is well scoped. I reproduced the bug and checked the change end-to-end against the built CLI (`nest build` with the `swc` builder, `@swc/core` 1.16.2), comparing the parent commit with this PR's head:

| `.swcrc` | parent (`bb93c826`) | this PR |
|---|---|---|
| none | exit 0, CJS | exit 0, CJS |
| `//` comment + trailing comma, `module.type: es6` | exit 0, **CJS** — bug reproduced | exit 0, **ESM** ✅ |
| invalid (`{ not valid json`) | exit 1 (SWC's own `failed to read .swcrc` error) | exit 1 (clearer Nest error) ✅ |
| empty file | exit 1 (SWC's own error) | exit 1 ✅ |
| **UTF-8 BOM** + valid JSON, `module.type: es6` | exit 0, CJS (silently ignored) | **exit 1** — `InvalidSymbol at offset 0` ❌ |

### Blocking

**1. A BOM-prefixed `.swcrc` now hard-fails the build, while SWC accepts it.** `jsonc-parser` reports `InvalidSymbol` at offset 0 for a leading `U+FEFF`; SWC strips it first (`s.trim_start_matches('\u{feff}')` — it's in the snippet quoted in the PR description). Since the goal here is to match SWC's parser, and this turns a build that exits 0 today into one that exits 1, it needs handling before merge. It's a one-liner — see the inline comment. Please add a test for it as well.

### Non-blocking

2. The `readFileSync` catch still maps *every* read error to "missing" (e.g. `EISDIR`/`EACCES` on the default `.swcrc` → silent `{}`), which is the same class of silent discard this PR sets out to remove. Inline suggestion.
3. `at offset N` is hard to act on in a multi-line config; a `line:column` is cheap to compute. Inline suggestion.
4. Tests: `process.chdir()` is new to this test suite and throws under Vitest's `threads` pool; `vi.spyOn(process, 'cwd')` is enough since the method only calls `process.cwd()`. Also `prettier --check` fails on the spec file (the `exitSpy` block). Inline comments.

### Verified, no action needed

- **Exiting on an invalid default `.swcrc` is not a breaking change in practice.** Nest passes `swcrc: true`, so SWC reads the same file itself and already aborts the build on invalid/empty content (rows 3–4 above). This PR only replaces SWC's error with a clearer one, consistent with what #3418 and #3332 did for `nest-cli.json` / `tsconfig`.
- **The new dependency adds no install footprint.** `jsonc-parser@3.3.1` is already a production transitive dependency (`@angular-devkit/core`, `@angular-devkit/schematics`, `@nestjs/schematics`) and dedupes to the same version; the lockfile diff is a single line. The named ESM imports resolve correctly in plain Node against the compiled output (its `main` is a UMD bundle without an `exports` map, so I checked outside of Vitest too).
- Full suite passes locally (87 files / 1040 tests), `tsc` build and `oxlint` are clean for the touched files.

### Prior art

I found no earlier issue/PR about `.swcrc` comments or swallowed parse errors, and no maintainer stance against this. The closest is #3041 (comments in `nest-cli.json`), where my objection was that a `.json` file should be valid JSON — that doesn't apply here, since `.swcrc` isn't a `.json` file and SWC itself defines the format as JSONC. The silent `{}` fallback dates back to the original SWC support (#2107) and was never discussed as a deliberate choice. #2343 even contains a user-pasted `.swcrc` with a trailing comma, i.e. a real-world instance of this trap.

### Possible follow-ups (out of scope)

- A non-object root (`42`, `"str"`, or SWC's array-of-configs form `[{ "test": ... }]`) flows straight into `deepMerge` and produces garbage options. Pre-existing with `JSON.parse`, but worth validating now that there's a proper error path.
- `lib/compiler/defaults/webpack-defaults.ts` still reads `tsconfig` with raw `JSON.parse`, which has the same comment/trailing-comma problem.

---
*Review prepared with Claude Code (Claude Fable 5.1). Every claim above was reproduced locally against the PR head `db1a55f0`: the touched spec and the full suite, a build of the CLI run against fixture projects (parent vs. PR), direct probes of `jsonc-parser` and `@swc/core`, and a throwaway spec for the suggested changes.*

lib/compiler/swc/swc-compiler.ts
stefanwaldhauser: Addressed in d1e908ab.

lib/compiler/swc/swc-compiler.ts
stefanwaldhauser: Addressed in d1e908ab.

lib/compiler/swc/swc-compiler.ts
stefanwaldhauser: Addressed in d1e908ab.

test/lib/compiler/swc/swc-compiler.spec.ts
stefanwaldhauser: Addressed in d1e908ab.

test/lib/compiler/swc/swc-compiler.spec.ts
stefanwaldhauser: Addressed in d1e908ab.

kamilmysliwiec:
lgtm

